PR #2, opened 2017-08-31
Add a benchmark for luacheck a pure Lua static analyzer 

fsfod:
This pull request adds the first large benchmark which uses luacheck a pure Lua static analyser to analyse the main source files of [ZeroBraneStudio ](https://github.com/pkulchenko/ZeroBraneStudio) a Lua based IDE concatenated together.
Currently the static analyst results are thrown away with out any processing since the various output formatter only write to the console or file I think.

Since this is first benchmark that uses external Lua modules worth mentioning a bit of how they work and how we could organize them.
Lua modules search locations is mainly based on the the environmental variable LUA_PATH and Lua variable package.path that contain path pattens with "?" substitution  location for the module names. If these variables are unmodified Lua will just search current working directory and totally oblivious to anything relative to a file calling require("modulename"). There two ways to handle this either update package.path for each benchmark run in the list to point to the directory the benchmark file is in and puts the Lua modules in there or have a shared location of all the different Lua modules and set package.path just once to that direct.


vext01:
> There two ways to handle this either update package.path for each benchmark run in the list to point to the directory the benchmark file is in and puts the Lua modules in there or have a shared location of all the different Lua modules and set package.path just once to that direct.

Can lua get the location of the current file? If so you could add to the package path relative to that?

benchmarks/luacheck/bench.lua
```diff
@@ -0,0 +1,17 @@
+local luacheck = require "luacheck" 
+
+local luafile = io.open("benchmarks/luacheck/ZeroBraneStudio.lua", "rb")
+local content = luafile:read("*all")
+luafile:close()
+
+function run_iter(count)
+
```
vext01: Lots of whitespace in this file.

benchmarks/luacheck/bench.lua
```diff
@@ -0,0 +1,17 @@
...
+    for i=1,count do
+        local start = os.clock()
```
vext01: Do we need this call to os.clock()?

benchmarks/luacheck/bench.lua
```diff
@@ -0,0 +1,17 @@
...
+        local report = luacheck.get_report(content)
+        
+        --luacheck.process_reports(report)
```
vext01: Commented code.

fsfod:
> Can lua get the location of the current file? If so you could add to the package path relative to that?
All you can get is the chunk name which is whatever relative or full path the Lua file was loaded with. There are examples of people trying to figure the path on stackoverflow kinda that way but it could be brittle to the working directory.

ltratt:
I don't have any strong opinions on the path thing. Edd's suggestion, if it's possible, puts the smallest burden on the user.

ltratt:
There are definitely cases where a program can't really work out where it's called from. Fortunately, from experience, these are rare enough that we can discount them. The basic trick, IME, is to canonicalise the path (the chunk path?), then do directory operations on that.

vext01:
I was wondering more if lua had a equivalent to `__file__`...

benchmarks/luacheck/bench.lua
fsfod: fixed

benchmarks/luacheck/bench.lua
fsfod: I have now setup a proper call to it but I haven't set any proper filters yet so it won't be doing much processing

LICENSE
```diff
@@ -0,0 +1,78 @@
+﻿--[[ ZeroBrane Studio License ]]-------------------------------------------
```
ltratt: I think we need a line / paragraph at the top that says something like "Our original code is licensed on a per-file basis. Other project's licenses are included below for reference."

ltratt:
OK, so where are we with the `__file__` equivalent thingy?

benchmarks/luacheck/bench.lua
fsfod: fixed

LICENSE
fsfod: I just used that

LICENSE
```diff
@@ -1,4 +1,6 @@
-﻿--[[ ZeroBrane Studio License ]]-------------------------------------------
+﻿Our original code is licensed on a per-file basis. Other project's licenses are included below for reference.
```
ltratt: Let's wrap this to 76 or 80 chars like the rest of the file.

benchmarks/luacheck_parser/bench.lua
```diff
@@ -0,0 +1,61 @@
...
+    local content = luafile:read("*all")
+    table.insert(sources, content)
+    luafile:close()   
+end
+
+   
+local parser = require "luacheck.parser"   
+    
+function run_iter(count)
+
```
ltratt: Extraneous blank space.

fsfod: This was already fixed in a later commit of the same push.

ltratt:
@vext01 are you OK with this?

vext01:
I guess so. It's so huge, it's kinda hard to review :)

I suppose a lot of it is external code anyway, so go ahead.

ltratt:
OK, so @fsfod if you squash, I'll merge.

ltratt:
[You'll also need to rebase everything against master, but that will probably happen anyway!]

fsfod:

ltratt:
A quick couple of typos I spotted in the commit messages: "combine there reports" -> "combine their reports"; "since its used" -> "since it's used". Feel free to do a force push to update them.

fsfod:
@ltratt ping. The commit messages are updated slightly.

ltratt:
Thanks!

PR #3, opened 2017-08-31
Basic Travis CI file

fsfod:
I have never really used travis before but this is my fist stab at a .travis.yml file.  I have only set it up to use a pre-compiled luajit package for quicker turn around time maybe we could use the matrix system to build and test each luajit mode separately. I also went with the more lightweight container mode with sudo: false instead of vm because I'm not doing anything fancy like make install idk if this could become an issues latter on if we also want run krun or something.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: c
+sudo: false
+
+before_script:
+  - sh ./build_luajit.sh
+
+script:
+  - ./builds/luajit_nojit simplerunner.lua
+  - ./builds/luajit simplerunner.lua
+  - ./builds/luajit_gc64 simplerunner.lua
+  - ./builds/luajit_dualnum simplerunner.lua


ltratt:
Before we can merge this, you'll need to set up Travis with this repo (it's a bit frustrating, but it is doable!). In order to test, it's often easiest to either a) push a new commit to this branch b) rebase and force push.

ltratt:
What does the `-` at the beginning of lines do? Suppress output?

fsfod:
It yaml's syntax. "List members are denoted by a leading hyphen (-) with one member per line"

ltratt:
Oh! I guess I've never used that before.

OK, I think it's time to squash, then I can merge.

fsfod:

ltratt:
Thanks!

PR #8, opened 2017-09-11
Start setting up krun

fsfod:
Add a krun file to run all the benchmarks with all the LuaJIT build flavours with placeholder iteration counts.

### DIFF
--- a/luajit.krun
+++ b/luajit.krun
@@ -1,0 +1,98 @@
+import os
+from krun.vm_defs import (LuaVMDef)
+from krun import EntryPoint
+from krun.util import fatal
+
+# Who to mail
+MAIL_TO = []
+
+# Maximum number of error emails to send per-run
+#MAX_MAILS = 2
+
+DIR = os.getcwd()
+
+VM_ENV = {
+  'LUA_PATH' : '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(os.path.join(DIR, "lualibs"))
+}
+
+HEAP_LIMIT = 2097152  # KiB
+STACK_LIMIT = 8192  # KiB
+
+VARIANTS = {
+    "default-lua": EntryPoint("bench.lua"),
+}
+
+ITERATIONS_ALL_VMS = 5
+
+VMS = {
+    'Normal': {
+        'vm_def': LuaVMDef("builds/luajit", VM_ENV),
+        'variants': ['default-lua'],
+        'n_iterations': ITERATIONS_ALL_VMS,
+    },
+    'GC64': {
+        'vm_def': LuaVMDef("builds/luajit_gc64", VM_ENV),
+        'variants': ['default-lua'],
+        'n_iterations': ITERATIONS_ALL_VMS,
+    },
+    'NoJIT': {
+        'vm_def': LuaVMDef("builds/luajit_nojit", env = VM_ENV),
+        'variants': ['default-lua'],
+        'n_iterations': ITERATIONS_ALL_VMS,
+    },
+    'DualNum': {
+        'vm_def': LuaVMDef("builds/luajit_dualnum", env = VM_ENV),
+        'variants': ['default-lua'],
+        'n_iterations': ITERATIONS_ALL_VMS,
+    },
+}
+
+BENCHMARKS = {
+    "binarytrees" : 1,
+    "nbody" : 1,
+    "fasta": 1,
+    "richards": 1,
+    "spectralnorm": 1,
+    "fannkuch_redux": 1,
+    "md5": 1,
+    "series": 1,
+    "luacheck_parser": 1,
+    "luacheck": 1,
+    "capnproto_encode": 1,
+    "capnproto_decode": 1,
+    "jsonlua_encode": 1,
+    "jsonlua_decode": 1,
+}
+
+# list of "bench:vm:variant"
+SKIP = [
+    #"*:Normal:*",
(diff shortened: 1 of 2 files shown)


luajit.krun
```diff
@@ -0,0 +1,96 @@
...
+    "series": 1,
+    "luacheck_parser": 1,
+    "luacheck": 1,
+    "capnproto_encode": 1,
+    "capnproto_decode": 1,
+    "jsonlua_encode": 1,
+    "jsonlua_decode": 1,
+}
+
+# list of "bench:vm:variant"
```
vext01: I'd recommend adding commented one-line-per-vm wildcard lines here so you can quickly disable VMs.

vext01: Also, add spaces `SKIP = [`. This was wrong in the Krun repo too! Oops.

luajit.krun
```diff
@@ -0,0 +1,96 @@
...
+SKIP=[
+    #"*:C:*",
+    #"*:PyPy:*",
+]
+
+N_EXECUTIONS = 1  # Number of fresh processes.
+
+# No. of seconds to wait before taking the initial temperature reading.
```
vext01: I'd kill this section. The default is probably OK.

luajit.krun
```diff
@@ -0,0 +1,96 @@
...
+    "jsonlua_decode": 1,
+}
+
+# list of "bench:vm:variant"
+SKIP=[
+    #"*:C:*",
+    #"*:PyPy:*",
+]
+
+N_EXECUTIONS = 1  # Number of fresh processes.
```
vext01: How many do you intend to run for the first round of the results?

luajit.krun
```diff
@@ -0,0 +1,96 @@
...
+#
+# Environment available for these commands:
+#   KRUN_RESULTS_FILE: path to results file.
+#   KRUN_LOG_FILE: path to log file.
+#   KRUN_ETA_DATUM: time the ETA was computed
+#   KRUN_ETA_VALUE: estimated time of completion
+#PRE_EXECUTION_CMDS = ["sudo service cron stop"]
+#POST_EXECUTION_CMDS = ["sudo service cron start"]
+
+# CPU pinning (off by default)
```
vext01: Are you sure you want pinning? Is LuaJIT multi-threaded?

luajit.krun
```diff
@@ -0,0 +1,96 @@
...
+ENABLE_PINNING = True
+
+# Lower and upper bound for acceptable APERF/MPERF ratios
+AMPERF_RATIO_BOUNDS = 0.995, 1.005
+
+# Rough rate of change in APERF per second above which a core is considered busy.
+# For many machines this is simply the base clock frequency, but generally
+# speaking, is undefined, so you need to check on a per-machine basis.
+AMPERF_BUSY_THRESHOLD = 3.4 * 1000 * 1000 * 1000 / 1000  # 3.4 GHz / 1000
```
vext01: I'd skip this section too until we get a proper benchmarking machine up and running.

ltratt:
Edd, I think this one is best done by you, so I've removed me from the PR.

ltratt:
Edd, I think this one is best done by you, so I've removed me from the PR.

luajit.krun
vext01: You can leave XXX markers as reminders if you like...

luajit.krun
fsfod: I thought it could switch around cores from context switches but maybe I misunderstanding what it does

vext01: You are right, that's what it's supposed to prevent, but you'd have to make sure luajit spins up an appropriate number of threads.

If you pin on a n-core machine, you will get a n-1 core "cset shield" to run on. You'd need to check that luajit doesn't spin up 4 cores worth of threads, as that will slow you down and distort your results (the jvm gets this wrong for example, so we gave up on pinning for the warmup experiment).

vext01: If you are going to use regular lua, you'd need to do a similar check for fairness.

fsfod: LuaJIT is not multi-threaded in anyway except its sampled based profiler.

ltratt: Let's not use pinning please. It opens up a new world of pain.

luajit.krun

luajit.krun

luajit.krun
fsfod: I don't really know what i should go with

vext01: I'd recommend 10 if you have time, for the initial "rough" results you will show cloudflare?

luajit.krun
vext01: If LuaJIT and the reference implementation are truly single threaded, then pinning would be OK. Does that change your opinion Laurie?

vext01:
I think we can kill the comments in the config file, as it's just duplication from the krun example.

fsfod:
> I think we can kill the comments in the config file, as it's just duplication from the krun example.

I'm not familiar with all options so having comments above the options helps a lot for me .

luajit.krun
ltratt: Nope :) I don't think we understand the effects of pinning well enough to make it worth the bother. One day, maybe we will know enough to be confident that turning it on is a good idea.

vext01: OK

vext01:
@fsod do you want to squash?

fsfod:
> @fsod do you want to squash?
@vext01 Done.

PR #9, opened 2017-09-12
Krun Tuning

fsfod:
Try to get one iteration of each benchmark to run for 0.1 seconds.

### DIFF
--- a/simplerunner.lua
+++ b/simplerunner.lua
@@ -18,7 +18,7 @@ local benchlist = {
     "luafun",
 }
 
-function runbench(name)
+function runbench(name, count)
     dofile("benchmarks/"..name.."/bench.lua")
 
     if not run_iter then
@@ -26,7 +26,7 @@ function runbench(name)
     end
 
     local start = os.clock()
-    run_iter(1)
+    run_iter(count or 1)
     print(name.." took", os.clock() - start)  
 
     --Clear the benchmark function so we know we're not running this benchmark again when we load another benchmark.
@@ -41,14 +41,14 @@ function run_all_benchmarks()
     end
 end
 
-function run_single_benchmarks(benchname)
+function run_single_benchmarks(benchname, count)
     assert(#benchlist ~= 0)
 
     benchname = benchname:lower()
     
     for i,name in ipairs(benchlist) do
         if name == benchname then
-            runbench(name)
+            runbench(name, count)
             return
         end
     end
@@ -57,7 +57,7 @@ function run_single_benchmarks(benchname)
 end
 
 if arg[1] then
-    run_single_benchmarks(arg[1])
+    run_single_benchmarks(arg[1], arg[2])
 else
     run_all_benchmarks()
 end
--- a/benchmarks/capnproto_encode/bench.lua
+++ b/benchmarks/capnproto_encode/bench.lua
@@ -4,7 +4,7 @@ local example_capnp = require "example_capnp"
 
 local list = {}
 
-for i=1, 100 do
+for i=1, 1000 do
     local data = {
         i0 = 32,
         i1 = 16,
--- a/benchmarks/jsonlua_encode/bench.lua
+++ b/benchmarks/jsonlua_encode/bench.lua
@@ -1,7 +1,7 @@
 local json = require "json"
 local list = {}
 
-for i=1, 100 do
+for i=1, 1000 do
     local data = {
         i0 = 32,
         i1 = 16,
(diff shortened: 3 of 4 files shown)


vext01:
So that's 0.1 seconds for which variant of the VM?

Is that on the machine you intend to run the preliminary results on?

Can you show me a Krun results file from running 1 process execution with say 10 in-process executions?

Any idea why git thinks the whole config file has changed?

fsfod:
The line endings got flipped when fixing a merge conflict I would of normally spotted that normally.

vext01:
OK, well if possible, can we push a fix, so I can see what really changed.

Thanks

fsfod:
i would have to force push if thats what you really want

vext01:
That's fine in this one instance.

vext01:
That's better. Now, how many in-process iterations did you run to calibrate. The first few iterations are bound to be super slow. You'd want to let it run a little before looking at the times.

fsfod:
I don't know what machine its going to run on i just had guess based on my linux machine time that is clocked higher than most servers but is a few generations old(sandy bridge 2600k).

fsfod:
I don't know what machine its going to run on i just had guess based on my linux machine time that is clocked higher than most servers but is a few generations old(sandy bridge 2600k).

simplerunner.lua
```diff
@@ -57,7 +57,7 @@ function run_single_benchmarks(benchname)
 end
 
 if arg[1] then
-    run_single_benchmarks(arg[1])
+    run_single_benchmarks(arg[1], arg[2])
```
vext01: Is this recursive?

vext01: ah i see, ignore me.

fsfod: what do you mean that's the optional run count it can be nil and default to 1 inside runbench

vext01: ignore this comment. I misread the code.

vext01:
Do you want to squash?

fsfod:
> Do you want to squash?
Do you have any comments on the JIT vs no JIT iteration times or anything?

vext01:
> Do you have any comments on the JIT vs no JIT iteration times or anything?

Well, I saw you had reduced the count. Presumably you experimented to find a suitable value?

fsfod:
it was just some guess, but here are the slow downs for no JIT. The extra time for one iteration and slow down amount
```
binarytrees        0.077231, 1.587435
capnproto_decode   0.333185, 3.374063
capnproto_encode   1.635851, 13.863944
fannkuch_redux     0.331934, 3.672644
fasta              0.213632, 2.698880
jsonlua_decode     0.184182, 2.335364
jsonlua_encode     0.013045, 1.099005
luacheck           -0.512662, 0.496973
luacheck_parser    0.026105, 1.183490
luafun             0.145883, 1.862993
md5                2.219446, 20.224372
nbody              0.985365, 7.003201
richards           0.348244, 3.225598
series             0.079451, 1.231476
spectralnorm       2.088578, 15.354350
```

vext01:
What are the two columns sorry? One figure is negative. Is that right?

vext01:
The second column "slow down amount", is how many times the jitted version?

vext01:
I think you should estimate now, as I did in mattermost, how long the experiment will take to run based on your above estimates.

fsfod:
> What are the two columns sorry? One figure is negative. Is that right?

Yes lua check runs faster with the jit off. The first is extra seconds a single no JIT iteration takes over the Normal JIT. The other is how many times slower no JIT vs JIT

using n_pexecs, n_iters = 10, 1500 but n_iters for NoJIT = 100. 
Times are based on the average time of one iteration of all the benchmarks added together for each VM. 
Time in hours.
```
Normal             13.224481
DualNum            13.562758
NoJIT              3.150929
GC64               13.835596
Total              43.773763
```

ltratt:
OK, so I'd like to know why there's a negative time here too:

```
luacheck           -0.512662, 0.496973
```

fsfod:
>yes luacheck runs faster with the jit off

ltratt:
Aha, I see.

vext01:
Well, your time estimate sounds good (In reality, krun will take a little longer to run these, as it has a few waits as part of normal operation).

If Laurie has no other comments, are we ready to squash?

fsfod:
@vext01  I had squashed most of the commits leaving what I think import changes left in the history. I also added back the pexce and iteration counts since you two are ok with the times

ltratt:
OK, I'll leave this to Edd to consider for merging.

vext01:
OK. Ready to merge?

fsfod:
@vext01 yes merge it.

PR #10, opened 2017-09-21
Add RaptorJIT

fsfod:
Add RaptorJIT to the VMs benchmarked.

### DIFF
--- a/build_luajit.sh
+++ b/build_luajit.sh
@@ -1,6 +1,5 @@
-#!/bin/sh
+#!/bin/bash
 #Build the various intresting flavors of LuaJIT
-
 
 if [ $# -gt 1 ]; then
     echo "Usage: build_luajit.sh [<source dir>]" 2>&1
@@ -21,6 +20,13 @@ else
     if [ ! -d "raptorjit_repo" ]; then 
       git clone https://github.com/raptorjit/raptorjit raptorjit_repo
     fi
+
+    if [ -f "commitlock.txt" ]; then 
+      source commitlock.txt
+      echo "Current luajit commit hash: ${luajit_commitid}"
+      echo "Current raptorjit commit hash: ${raptorjit_commitid}"
+      (cd luajit_repo && git checkout ${luajit_commitid})
+      (cd raptorjit_repo && git checkout ${raptorjit_commitid})
     fi
 fi
 
--- a/build_luajit.sh
+++ b/build_luajit.sh
@@ -17,11 +17,18 @@ else
     if [ ! -d "luajit_repo" ]; then 
       git clone -b v2.1 https://github.com/LuaJIT/LuaJIT luajit_repo
     fi
+
+    if [ ! -d "raptorjit_repo" ]; then 
+      git clone https://github.com/raptorjit/raptorjit raptorjit_repo
+    fi
+    fi
 fi
 
 ljbins=${ljbins:=builds}
 
 mkdir -p ${ljbins}
+# Make directory path relative so we don't have to worry about what path we have to pass to make.
+ljbins="$( cd "$ljbins" && pwd )"
 
 #Unmodified build with 32 bit sized gc object pointers. Object allocataion limited to the lower 4gb virtual address space
 make -C ${ljsrc} clean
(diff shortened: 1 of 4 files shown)


build_luajit.sh
```diff
@@ -17,6 +17,10 @@ else
     if [ ! -d "luajit_repo" ]; then 
       git clone -b v2.1 https://github.com/LuaJIT/LuaJIT luajit_repo
     fi
+    
+    if [ ! -d "ratorjit_repo" ]; then 
+      git clone https://github.com/raptorjit/raptorjit ratorjit_repo
```
vext01: Should we be using a fixed tag/hash for our VMs?

luajit.krun
```diff
@@ -46,6 +46,11 @@ VMS = {
         'variants': ['default-lua'],
         'n_iterations': ITERATIONS_ALL_VMS,
     },
+    'RaptorJIT': {
+        'vm_def': LuaVMDef("builds/raptorjit", env = VM_ENV),
```
vext01: In Python, typically kwargs (keyword args) don't use spaces around `=`. I know, it's inconsistent, but that's the way the Python folk do it :P

luajit.krun
```diff
@@ -69,6 +74,7 @@ BENCHMARKS = {
 # list of "bench:vm:variant"
 SKIP = [
     "capnproto_decode:GC64:*", # Sometime crashes with GC64
+    "capnproto_decode:RaptorJIT:*"
```
vext01: Does this crash too?

build_luajit.sh
ltratt: Good point -- yes, we should.

Tom: what Edd means is that by fixing a hash, we know precisely which version of each JIT we're benchmarking.

That said, we could allow people to specific an option that means "just download the latest versions".

vext01:
Travis choked!

luajit.krun
fsfod: It seems like it even though I'm not building it as GC64

ltratt: What sort of crash are you seeing? We should probably report that upstream to raptorjit.

fsfod:
He killed off the bootstrap support using a single Puc Lua file. so you have to have LuaJIT installed to run dynasm and stuff.

vext01:
You can add packages inside travis if that helps.

luajit.krun
vext01: If it's crashing, let's add a comment saying that.

fsfod:
That was what i was doing but i could just reuse the LuaJIT we already built.

vext01:
That's also fine :)

build_luajit.sh
```diff
@@ -26,6 +26,8 @@ fi
 ljbins=${ljbins:=builds}
 
 mkdir -p ${ljbins}
+# Make directory path relative so we don't have to worry about what path we have to pass to make.
+ljbins="$( cd "$ljbins" && pwd )"
```
vext01: `${ljbins}` for consistnecy?

What does `$(...)` do here? Never seen this before.

build_luajit.sh
fsfod: I added locking to a commit hash

build_luajit.sh
```diff
@@ -21,6 +20,13 @@ else
     if [ ! -d "raptorjit_repo" ]; then 
       git clone https://github.com/raptorjit/raptorjit raptorjit_repo
     fi
+
+    if [ -f "commitlock.txt" ]; then 
+      source commitlock.txt
+      echo "Current luajit commit hash: ${luajit_commitid}"
+      echo "Current raptorjit commit hash: ${raptorjit_commitid}"
+      (cd luajit_repo && git checkout ${luajit_commitid})
```
ltratt: The `(...)` brackets aren't necessary.

build_luajit.sh
fsfod: I assume capture the stdout of the command run inside it.

build_luajit.sh
fsfod: I have to change the directory back if i don't run it like that

luajit.krun

build_luajit.sh
ltratt: Oh, interesting -- I didn't realise it had that effect!

build_luajit.sh
vext01: Ah subshell. Fair enough.

I always use backticks, but this is fine too.

build_luajit.sh
vext01: today I learned!

vext01:
What is the status of this @fsfod? Looks like you have addressed all of the comments?

fsfod:
@ vext01 The only thing I didn't do was add the commitlock file or an opt out out for it but maybe that can be done when we pick a fixed target.

vext01:
Up to you. I don't know the purpose of the commitlock file.

I'm OK to merge this as long as your next PR fixes the VM versions.

Do you want to squash?

fsfod:
The commitlock will contain the commit hash ids to checkout when building LuaJIT and RaptorJIT from source with build_luajit.sh

vext01:
OK, so that seems more like a change for the next PR then? What do you think?

fsfod:
Yeah lets just squash.

vext01:
Done?

fsfod:
@vext01 yes.

vext01:
OK. So next let's fix those VM versions.

PR #11, opened 2017-09-27
Add a JIT stats tracking library 

fsfod:
Add jitstats a pure Lua low overhead JIT stats tracking library that use the vmevent hooks to track trace behaviour and builds the stats based on trace start\stop\aborts .

### DIFF
--- a/jitstats.lua
+++ b/jitstats.lua
@@ -1,0 +1,448 @@
+local jit = require("jit")
+local bc = require("jit.bc")
+local vmdef = require("jit.vmdef")
+local jutil = jit.util or require("jit.util")
+local funcinfo, funcbc = jutil.funcinfo, jutil.funcbc
+local band = bit.band
+require("table.clear")
+require("table.new")
+
+local bcnames, bcname_lookup = {}, {} 
+local bccount = 0
+for i = 0, #vmdef.bcnames-1, 6 do
+    local name = string.sub(vmdef.bcnames, i, i+6):match "^%s*(.-)%s*$"
+    bcnames[bccount] = name
+    bcname_lookup[name] = bccount
+    bccount = bccount + 1
+end
+
+local traceerr_lookup = {}
+for i, msg in pairs(vmdef.traceerr) do
+    traceerr_lookup[msg] = i
+end
+
+local traceerror_info = {
+  {name = "RECERR",  msg = "error thrown or hook called during recording"},
+  {name = "TRACEUV", msg = "trace too short"},
+  {name = "TRACEOV", msg = "trace too long"},
+  {name = "STACKOV", msg = "trace too deep"},
+  {name = "SNAPOV",  msg = "too many snapshots"},
+  {name = "BLACKL",  msg = "blacklisted", displaymsg = "Function or loop blacklisted"},
+  {name = "RETRY",   msg = "retry recording"},
+  {name = "NYIBC",   msg = "NYI: bytecode %d", displaymsg = "NYI: bytecode"},
+  
+  -- Recording loop ops.
+  {name = "LLEAVE",  msg = "leaving loop in root trace"},
+  {name = "LINNER",  msg = "inner loop in root trace"},
+  {name = "LUNROLL", msg = "loop unroll limit reached"},
+  
+  -- Recording calls/returns.
+  {name = "BADTYPE", msg = "bad argument type"},
+  {name = "CJITOFF", msg = "JIT compilation disabled for function"},
+  {name = "CUNROLL", msg = "call unroll limit reached"},
+  {name = "DOWNREC", msg = "down-recursion, restarting"},
+  {name = "NYIFFU",  msg = "NYI: unsupported variant of FastFunc %s", displaymsg = "NYI: unsupported variant of FastFunc"},
+  {name = "NYIRETL", msg = "NYI: return to lower frame"},
+  
+  -- Recording indexed load/store.
+  {name = "STORENN", msg = "store with nil or NaN key"},
+  {name = "NOMM",    msg = "missing metamethod"},
+  {name = "IDXLOOP", msg = "looping index lookup"},
+  {name = "NYITMIX", msg = "NYI: mixed sparse/dense table"},
+  
+  -- Recording C data operations. */
+  {name = "NOCACHE", msg = "symbol not in cache"},
+  {name = "NYICONV", msg = "NYI: unsupported C type conversion"},
+  {name = "NYICALL", msg = "NYI: unsupported C function type"},
+  
+  -- Optimizations.
+  {name = "GFAIL",   msg = "guard would always fail"},
+  {name = "PHIOV",   msg = "too many PHIs"},
+  {name = "TYPEINS", msg = "persistent type instability"},
+  
+  -- Assembler.
+  {name = "MCODEAL", msg = "failed to allocate mcode memory"},
+  {name = "MCODEOV", msg = "machine code too long"},
+  {name = "MCODELM", msg = "hit mcode limit (retrying)"},
+  {name = "SPILLOV", msg = "too many spill slots"},
+  {name = "BADRA",   msg = "inconsistent register allocation"},
+  {name = "NYIIR",   msg = "NYI: cannot assemble IR instruction %d"},
+  {name = "NYIPHI",  msg = "NYI: PHI shuffling too complex"},
+  {name = "NYICOAL", msg = "NYI: register coalescing too complex"},
+}
+
+local traceerr = {}
+local traceerr_fromid = table.new(#vmdef.traceerr, 0)
+local traceerr_displaymsg = {}
+
+for i, evt in ipairs(traceerror_info) do
+    local id = traceerr_lookup[evt.msg]
+    if id then
+        traceerr[evt.name] = id
+        traceerr_fromid[id] = evt.name
+    end
+    traceerr_displaymsg[evt.name] = evt.displaymsg or evt.msg
+end
+
+local function printf(fmt, ...)
+    print(string.format(fmt, ...))
+end
+
+local debugprint = function () end
+
+local function fmtfunc(func, pc)
+    local fi = funcinfo(func, pc)
+    if fi.loc then
+        return fi.loc
+    elseif fi.ffid then
+        return vmdef.ffnames[fi.ffid]
+    elseif fi.addr then
+        return string.format("C:%x", fi.addr)
+    else
+        return "(?)"
+    end
+end
+
+-- Format trace error message.
+local function fmterr(err, info)
+    if type(err) == "number" then
+        if type(info) == "function" then
+          info = fmtfunc(info) 
+        end
+        err = string.format(vmdef.traceerr[err], info)
+    end
+    return err
+end
+
+local jitstats = {
+    isrunning = false,
+    debug = false,
+    stats = {}
+}
+
+local stats = jitstats.stats
+
+local vmevent = {}
+
+local curr_parentid, curr_stitched
+-- Trace started
+function vmevent.start(tr, func, pc, parentid, exitnum)
+    curr_parentid = parentid
+    curr_stitched = parentid and parentid > 0 and exitnum == -1
+
+    debugprint("START(%d): parent = %d", tr, parentid or -1)
+    
+    stats.starts = stats.starts + 1
+    if curr_stitched then
+        stats.stitch_starts = stats.stitch_starts + 1
+    elseif parentid then
+        stats.side_starts = stats.side_starts + 1
+    end 
+end
+
+--Trace Stopped
+function vmevent.stop(tr)
+    local info = jutil.traceinfo(tr)
+     
+    debugprint("STOP(%d): link = %s", tr, info.linktype)
+    
+    stats.completed = stats.completed + 1
+    
+    if curr_stitched then
+        stats.stitch_completed = stats.stitch_completed + 1
+    elseif curr_parentid then
+        stats.side_completed = stats.side_completed + 1
+    end
+
+    stats.linktypes[info.linktype] = (stats.linktypes[info.linktype] or 0) + 1
+end
+
+--Trace Aborted
+function vmevent.abort(tr, func, pc, code, errinfo)
+    stats.aborts = stats.aborts + 1
+    
+    local reason = ""
+    if jitstats.debug then
+        reason = fmterr(code, errinfo)
+        reason = reason:gsub("bytecode (%d+)", function(c)
+            c = tonumber(c)
+            assert(bcnames[c], c)
+            return "bytecode "..bcnames[c]
+        end)
+    end
+
+    if curr_stitched then
+        stats.stitch_aborts = stats.stitch_aborts + 1
+    elseif curr_parentid then
+        debugprint("ABORT(%d): parent = %d, %s", tr, curr_parentid, reason)
+        stats.side_aborts = stats.side_aborts + 1
+    else
+        debugprint("ABORT(%d): %s", tr, reason)
+    end
+
+    -- We sometimes don't get a bc for the stop location
+    local bc = funcbc(func, pc)
+    local opcode = bc and band(bc, 0xff)
+    
+    if code == traceerr.LLEAVE then
+        debugprint("LLEAVE: %d/%s", opcode, bcnames[opcode])
+    elseif code == traceerr.LUNROLL then
+        
+    elseif code == traceerr.NYIBC then
+        bcname = bcnames[errinfo]
+        stats.nyi_bc[bcname] = (stats.nyi_bc[bcname] or 0) + 1
+    end
+
+    -- Use internal enum name of the trace abort error if possible
+    local tracerror = traceerr_fromid[code] or vmdef.traceerr[code]
+    stats.abort_counts[tracerror] = (stats.abort_counts[tracerror] or 0) + 1
+end
+
+function vmevent.flush()
+    debugprint("Full trace flush")
+    stats.traceflush = stats.traceflush + 1
+end
+
+function vmevent.texit(id, exitno)
+    debugprint("EXIT(%d): exit = %d", id, exitno)
+    stats.exits = stats.exits + 1
+end
+
+local function vmeventcb(what, ...)
+    local cb = vmevent[what]
+    if cb then 
+        cb(...)
+    end
+end
+
+function jitstats.start()
+    if jitstats.isrunning then
+        return false
+    end
+
+    jit.attach(vmeventcb, "trace")
+    jit.attach(vmevent.texit, "texit")
+    jitstats.isrunning = true
+    return true
+end
+
+function jitstats.stop()
+    if not jitstats.isrunning then
+        return false
+    end
+
+    jit.attach(vmeventcb)
+    jit.attach(vmevent.texit)
+    jitstats.isrunning = false
+    return true
+end
+
+local function initorclear(tab, key)
+    local t = tab[key]
+    if t then
+        table.clear(t)
+    else
+        tab[key] = {}
+    end
+end
+
+local function reset_subtables(statstbl)
+    initorclear(statstbl, "abort_counts")
+    initorclear(statstbl, "nyi_bc")
+    initorclear(statstbl, "linktypes")
+end
+
+function jitstats.reset(statstbl)
+    statstbl = statstbl or stats
+    statstbl.starts = 0
+    statstbl.aborts = 0
+    statstbl.completed = 0
+    statstbl.stitch_starts = 0
+    statstbl.stitch_aborts = 0
+    statstbl.stitch_completed = 0
+    statstbl.side_starts = 0
+    statstbl.side_aborts = 0  
+    statstbl.side_completed = 0
+    statstbl.exits = 0
+    statstbl.traceflush = 0
+    --Clear or initlize 
+    reset_subtables(statstbl)
+end
+
+-- Note no attempt is made to handle circular references
+local function table_copy(src, dst)
+    dst = dst or {}
+    for k,v in pairs(src) do
+        if type(v) == "table" then
+            dst[k] = table_copy(v, dst[k])
+        else
+            dst[k] = v
+        end
+    end
+    return dst
+end
+
+--Optionally pass in a table to write the snapshot to instead of creating a new table
+function jitstats.getsnapshot(dst)
+    if dst then
+        reset_subtables(dst)
+    end
+    return table_copy(stats, dst)
+end
+
+local function diffstatbl(old, new, result, key)
+    assert(old[key], "missing old sub table")
+    assert(new[key], "missing new sub table")
+    assert(result[key], "missing result sub table")
+    old = old[key]
+    new = new[key]
+    result = result[key]
+
+    for k, v in pairs(new) do
+        local change = v - (old[k] or 0)
+        if change ~= 0 then
+            result[k] = change
+        end
+    end
+    return result
+end
+
+-- Get the diff between two jit stat snapshots
+function jitstats.diffsnapshots(snap1, snap2)
+    local result = {
+        abort_counts = {},
+        nyi_bc = {},
+        linktypes = {},
+    }
+
+    for k, v in pairs(snap2) do   
+        if type(v) == "number" then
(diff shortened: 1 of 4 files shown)


ltratt:
Before we start off the review, can you give more details about a) how to use this b) what the output looks like?

fsfod:
A basic use looks like this.

``` lua
local jitstats = require("jitstats")
-- Starts recording by hooking the vmevents
jitstats.start()
-- Lets make some traces
for i=1, 200 do
  if i < 100 then
    a = i 
  else
    b = i
  end
end
jitstats.print()
-- Stops recording by unhooking the vmevents
jitstats.stop()
```

An example of the output looks like this which was generated from one iteration of the the luacheck benchmark:

```
Trace Stats:
  Started 28060, side 18801(67%), stitch 908
  Aborted 12787(45%), stitch 101, side 6388
    1 error thrown or hook called during recording
    3629 NYI: bytecode %d
    1502 leaving loop in root trace
    3545 inner loop in root trace
    1914 loop unroll limit reached
    91 call unroll limit reached
    2026 down-recursion, restarting
    49 persistent type instability
    30 failed to allocate mcode memory
  NYI BC Aborts
    1803 ISNEXT
    2 FNEW
    531 ITERN
    1293 VARG
 Uncompiled exits taken 178235
```

ltratt:
@fsfod I notice you're pushing commits -- is this ready for review yet?

fsfod:
@ltratt  Theres no output for krun yet but I have stats being printed out as json to the console working that you can see the in travis.

ltratt:
@fsfod I don't mean the Krun integration (I guess that's a separate PR). What I meant is: you're still pushing commits, so can Javier review this? If so, I would guess you would need to squash and say "this is now ready for review".

ltratt:
@fsfod So it's not ready for review. When it is, please squash it and then say "ready for review" so that I know and can invite Javier. Thanks!

fsfod:
@ltratt ready for review.

ltratt:
I've assigned @javierguerragiraldez -- thanks Javier :) I will also do a review (though it'll probably be somewhat shallow).

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+  {name = "MCODEOV", msg = "machine code too long"},
+  {name = "MCODELM", msg = "hit mcode limit (retrying)"},
+  {name = "SPILLOV", msg = "too many spill slots"},
+  {name = "BADRA",   msg = "inconsistent register allocation"},
+  {name = "NYIIR",   msg = "NYI: cannot assemble IR instruction %d"},
+  {name = "NYIPHI",  msg = "NYI: PHI shuffling too complex"},
+  {name = "NYICOAL", msg = "NYI: register coalescing too complex"},
+}
+
+local traceerr, traceerr_fromid = {}, table.new(#vmdef.traceerr, 0)
```
ltratt: Wouldn't this be more readable as two assignments on separate lines?

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+    if select("#", ...) == 0 then
+        print(fmt)
+    else
+        print(string.format(fmt, ...))
+    end
+end
+
+local debugprint = function () end
+
+-- Stolen from dump.lua
```
ltratt: What's the license of `dump.lua`?

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+    local bc = funcbc(start_func, start_pc)
+    local bcname = bc and bcnames[band(bc, 0xff)]
+    if bcname == "IFUNCF" then
+        print"blacklist"
+        stats.func_blacklisted = stats.func_blacklisted + 1
+    elseif bcname == "ILOOP" or bcname == "IFORL" or bcname == "IITERL" then
+        stats.loop_blacklisted = stats.loop_blacklisted + 1
+    end
+
+    -- We sometimes don't can't find bc
```
ltratt: I can't parse the sentence: don't or can't?

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+    end
+    return dst
+end
+
+local function initorclear(tab, key)
+    tab[key] = tab[key] or {}
+    table.clear(tab[key])
+end
+
+--Optionally pass in a table to copy to snapshot to instead of creating a new table
```
ltratt: Too many "to"s.

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+    end
+    printf(" Uncompiled exits taken %d", statstbl.exits)
+
+    if running then
+        jitstats.start()
+    end
+end
+
+-- Turn off the jit for all the functions in this file so we don't contribute to the stats when being called. Sadly 
+-- the the effect of this call is defered with respect to switching the function header bytecode to interpreter only
```
ltratt: "the the"

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+    printf(" Uncompiled exits taken %d", statstbl.exits)
+
+    if running then
+        jitstats.start()
+    end
+end
+
+-- Turn off the jit for all the functions in this file so we don't contribute to the stats when being called. Sadly 
+-- the the effect of this call is defered with respect to switching the function header bytecode to interpreter only 
+-- mode to the first  time each function trys to start a trace.
```
ltratt: "first__time" (two spaces).

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+        printf(" Loops blacklisted: %d", statstbl.loop_blacklisted)
+    end
+    printf(" Uncompiled exits taken %d", statstbl.exits)
+
+    if running then
+        jitstats.start()
+    end
+end
+
+-- Turn off the jit for all the functions in this file so we don't contribute to the stats when being called. Sadly
```
ltratt: This comment is a really long sentence and I can't quite parse the last part of it. Maybe break it up into two sentences?

ltratt:
A dumb question: can you call `jitstats.print()` after `jitstats.stop()`, or must it be called before?

fsfod:
jitstats.stop() only unhooks the vm events were listening to all the stat data stays unchanged. jitstats.print also defensively does a temp jitstats.stop to really make sure we don't add extra stats from own code running in jitstats.print.

ltratt:
If `print()` calls `stop()` does it also call `start()`? Otherwise I'm a bit confused by the example you pasted earlier which ends:

```lua
jitstats.print()
-- Stops recording by unhooking the vmevents
jitstats.stop()
```

fsfod:
Ses it call start at the end. This is what it does

```
function jitstats.print(statstbl, msg)
    local running = jitstats.stop()
...
    if running then
        jitstats.start()
    end
end
```

ltratt:
Right, got it.

jitstats.lua
fsfod: Its not actually my comment https://github.com/geoffleyland/luatrace/blob/master/lua/jit/annotate.lua#L15 which is MIT licensed dump.lua is actually part of LuaJIT.

ltratt: If it's part of luajit, why don't we import it?

fsfod: All those functions are internal to that module they currently look like this in dump.lua
```
local function fmtfunc(func, pc)
  local fi = funcinfo(func, pc)
  if fi.loc then
    return fi.loc
  elseif fi.ffid then
    return vmdef.ffnames[fi.ffid]
  elseif fi.addr then
    return format("C:%x", fi.addr)
  else
    return "(?)"
  end
end

-- Format trace error message.
local function fmterr(err, info)
  if type(err) == "number" then
    if type(info) == "function" then info = fmtfunc(info) end
    err = format(vmdef.traceerr[err], info)
  end
  return err
end

File: jitstats.lua [hunk 8]
```diff
@@ -0,0 +1,462 @@
...
+
+local bccount = 0
+
+for i = 0, #vmdef.bcnames-1, 6 do
+    local name = string.sub(vmdef.bcnames, i, i+6):match "^%s*(.-)%s*$"
+    bcnames[bccount] = name
+    bcname_lookup[name] = bccount
+    bccount = bccount + 1
+end
+
```
javierguerragiraldez: limit `bccount` scope?

jitstats.lua
javierguerragiraldez: the messages aren't too long, i'd just loose the `name`/`msg` keys, like:
```
local traceerror_info = {
  {name = "RECERR",  msg = "error thrown or hook called during recording"},
  {name = "TRACEUV", msg = "trace too short"},
  {name = "TRACEOV", msg = "trace too long"},
```

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+local TRERR_BLACKL, TRERR_LLEAVE, TRERR_LUNROLL, TRERR_NYIBC = traceerr.BLACKL, traceerr.LLEAVE, traceerr.LUNROLL, traceerr.NYIBC
+
+local function printf(fmt, ...)
+    if select("#", ...) == 0 then
+        print(fmt)
+    else
+        print(string.format(fmt, ...))
+    end
+end
```
javierguerragiraldez: no need for the condition; `string.format(s)` is valid.  my typical `printf()` looks like
```
local function printf(fmt, ...)
    io.write(fmt:format(...))
end
```
(note it doesn't add a newline, just like in C)
but can be as simple as:
```
local function printf(...)
   print(string.format(...))
end
```

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+    debug = false,
+    stats = {}
+}
+
+local stats = jitstats.stats
+
+local vmevent = {}
+
+local curr_parentid, curr_stitched
+local start_pc, start_func
```
javierguerragiraldez: these variables are used only on `vmevent.start/stop/abort()`, wrap in a `do...end` to limit scope?

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+    end
+
+    stats.linktypes[info.linktype] = (stats.linktypes[info.linktype] or 0) + 1
+end
+
+--Trace Aborted
+function vmevent.abort(tr, func, pc, code, errinfo)
+    stats.aborts = stats.aborts + 1
+    
+    local reason
```
javierguerragiraldez: don't leave uninitialized.  it doesn't matter right now, but a change in the `debugprint()` behaviour could fail.  IOW: non-local invariant violation.

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+    dst = dst or {}
+    for k,v in pairs(src) do
+        if type(v) ~= "table" then
+            dst[k] = v
+        else
+            dst[k] = table_copy(v, dst[k])
+        end
+    end
+    return dst
+end
```
javierguerragiraldez: - maybe a warning about not being safe with self-referenting tables?
- on full `if...then...else...end` constructs, I typically find more readable to put the "positive" condition in as `then`, and the negative in the `else`.  like:
```
if type(v) == 'table' then
   dst[k] = table_copy(v, dst[k])
else
   dst[k] = v
end
```
- what about this form? (untested!):
```
local function table_copy(src, dst)
    if type(src) ~= 'table' then return src end
    dst = dst or {}
    for k, v in pairs(src) do
        dst[k] = table_copy(v, dst[k])
    end
    return dst
end
```

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+        initorclear(dst, "linktypes")
+    end
+
+    return table_copy(stats, dst)
+end
+
+local function diffstatbl(old, new, result, key)
+    assert(old[key], "missing old sub table")
+    assert(new[key], "missing new sub table")
+    assert(result[key], "missing old sub table")
```
javierguerragiraldez: these asserts can't fire in the current use.  while programming defensively is usually good, on short functions that are used on only one way it can obscure the intention.
two alternatives:
- just get rid of it.  reduce the scope of the function so there's no temptation to use somewhere else where the contract isn't followed.
- change the interface, removing the `key` parmenter.  something like this (untested!):
```
do
	local function diffstatbl(old, new)
		if not new then return old end -- or table_copy(old)  ?
		if type(new) == 'number' then
			return v - (old or 0)
		end
		local r = {}
		for k, v in pairs(new) do
			r[k] = v - (old[k] or 0)
		end
		return r
	end
	
	function jitstats.diffsnapshots(snap1, snap2)
		return {
			abort_counts = diffstatbl(snap1.abort_counts, snap2.abort_counts)
			nyi_bc = diffstatbl(snap1.nyi_bc, snap2.nyi_bc)
			linktypes = diffstatbl(snap1.linktypes, snap2.linktypes)
		}
	end
end
```

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+        return false
+    end
+
+    jit.attach(vmeventcb)
+    jit.attach(vmevent.texit)
+    jitstats.isrunning = false
+    return true
+end
+
+local last_starts, last_exits
```
javierguerragiraldez: shadowed by new locals further down

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+    end
+
+    diffstatbl(snap1, snap2, result, "abort_counts")
+    diffstatbl(snap1, snap2, result, "nyi_bc")
+    diffstatbl(snap1, snap2, result, "linktypes")
+
+    return result
+end
+
+local last_starts, last_exits
```
javierguerragiraldez: these look like private vars.  reduce scope?

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+    return true
+end
+
+function jitstats.setdebugprint(enable)
+    jitstats.debug = enable
+    debugprint = (enable and printf) or function() end
+end
+
+local function topercent(total, subval)
+    return math.floor(subval / total * 100)
```
javierguerragiraldez: `math.floor(.... +0.5)` to round correctly

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+end
+
+local function default_sorter(t, k1, k2)
+    return t[k1] > t[k2] 
+end
+
+local function table_sortkeys(t, sorter)
+    local keys = table_keys(t) 
+    sorter = sorter or default_sorter
+    table.sort(keys, function(k1, k2) return sorter(t, k1, k2) end)
```
javierguerragiraldez: YAGNI?  just do `table.sort(keys, function(k1, k2) return t[k1]>t[k2] end)` and remove the `sorter` parameter

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+  {name = "SPILLOV", msg = "too many spill slots"},
+  {name = "BADRA",   msg = "inconsistent register allocation"},
+  {name = "NYIIR",   msg = "NYI: cannot assemble IR instruction %d"},
+  {name = "NYIPHI",  msg = "NYI: PHI shuffling too complex"},
+  {name = "NYICOAL", msg = "NYI: register coalescing too complex"},
+}
+
+local traceerr, traceerr_fromid = {}, table.new(#vmdef.traceerr, 0)
+
+for i, evt in pairs(traceerror_info) do
```
javierguerragiraldez: `ipairs`?

jitstats.lua
```diff
@@ -0,0 +1,462 @@
...
+        statstbl = stats
+    else
+        assert(type(statstbl) == "table", type(statstbl))
+        msg = "Trace Stats:"
+    end
+
+    printf(msg)
+    printf("  Started %d, side %d(%g%%), stitch %d", statstbl.starts, statstbl.side_starts, topercent(statstbl.starts, statstbl.side_starts), statstbl.stitch_starts)
+    printf("  Completed Link types: %s", statstbl_concat(statstbl.linktypes))
+    printf("  Aborted %d(%g%%), side %d, stitch %d", statstbl.aborts, topercent(statstbl.starts, statstbl.aborts), statstbl.side_aborts, statstbl.stitch_aborts)
```
javierguerragiraldez: this is hard to maintain.  what about a simplistic template?  even just like this (untested!):
```
print(([[
   Started {starts}, side {side_starts}({pct_sides}%), ...
   ...
]]):gsub('{([_%w]+)}', copy_table(statstbl, {
    pct_sides = topercent(statstbl.starts, statstbl.side_starts),
    ...
})
```

jitstats.lua
fsfod: They did fire for me before because I didn't assign one of the tables in another function. I also did have it without the key parameter before but it felt error prone and verbose to type out the field 3 times when adding a new sub table.

javierguerragiraldez: yes, it's mostly a style preference, and what anybody considers more readable vs. safe.  in my opinion, if a function is just a helper, used in a single place, very close to where it's defined, and with the scope clearly limited, then safety can be relaxed a bit.

if it were a generic utility, or (much more!) an exported function, then yes, do it very defensively!

but even in that case, if the result of the user not following the contract (i.e. giving a string instead of a table) is a simple and obvious error (i.e. `"attempt to index local 's' (a string value)"`), then i wouldn't be worried.

jitstats.lua
fsfod: fixed

jitstats.lua
fsfod: fixed

jitstats.lua
fsfod: fixed

jitstats.lua
fsfod: fixed

jitstats.lua
fsfod: fixed

jitstats.lua
fsfod: fixed

jitstats.lua
fsfod: fixed

jitstats.lua
fsfod: fixed

jitstats.lua
fsfod: Now removed since function they were used in function didn't make sense was redesigned not to use them.

jitstats.lua
fsfod: I was probably using it with non sting values in the past but i was also thinking of letting the user supply there own output function as well

jitstats.lua
fsfod: I was thinking before of making topercent actually returning a string format of %d(%g%%) to simplify things. I have made my own poor man’s moustache before https://github.com/fsfod/LuaJIT/blob/gcperflog/src/perflog_apigen.lua#L43

javierguerragiraldez: haven't we all?  (tip: use a capture group and you can lose the `key:sub(3,-3)`, and the whole function is just a table)
but the thing with these "lots of printf()s" blocks is that when you add data fields or change the phrasing, you start to loose count of which %d is which value.  that's when a `:gsub({{...}}, values)` comes handy

jitstats.lua
fsfod: fixed

jitstats.lua
fsfod: I just went with the most common one first.

jitstats.lua
fsfod: I've never really needed to scope variables like that

jitstats.lua
fsfod: I think you two are talking about different lines. I used named fields because I was thinking of adding more like a better display string without all the printf garbage.

ltratt: Yes, I'm not sure we're talking about the same lines.

ltratt:
High-level question: should we make this (now? later?) a stand-alone library? And we should probably add some basic documentation so that other people can use it. I don't know what the standard Lua way for documenting things is, but, assuming there is such a standard, we should probably use that.

fsfod:
It is standalone except for depending on LuaJITs own internal Lua modules.

ltratt:
OK, so should it be in a stand-alone repository? I think the answer is probably "yes", but I'm open-minded. [I don't know how Lua distributes packages.]

fsfod:
They use https://luarocks.org/ but i think its bit early for that.

jitstats.lua
fsfod: I was going to use it to sort by key as well but idk if there would be much use in that.

fsfod: fixed i just simplified it for its single use

jitstats.lua
fsfod: fixed with a template system for the main unconditional block of text

jitstats.lua
fsfod: how about this @ltratt
> Turn off the jit for all the functions in this file so they don't try JIT. Sadly the effect
> of this call is deferred with respect to switching the function header bytecode. It is only set 
> to interpreter only mode the first time each function tries to start a trace.

ltratt: Let's capitalise "JIT" consistently. Unfortunately I still don't understand sentences 2 and 3 :(

ltratt:
OK, so once that documentation nit I pointed out is fixed, is this ready for Javier to do another review round?

fsfod:
I think i have responded\fixed to most of what he commented on

javierguerragiraldez:
> I think i have responded\fixed to most of what he commented on

yes.  i've been watching the new commits.  i'll try to do a test run of the code later today

fsfod:
I never really explained the snapshot api but maybe @javierguerragiraldez  figured most of it and that most of the other api's like reset and print are designed to also accept being passed a stats snapshot. The one thing i was thinking of adding was api to switch out the current stats state table.

fsfod:
This is an example of how i was using it in simple runner

``` lua
local start, stop = {}, {}
jitstats.getsnapshot(start)
run_iter(1)
jitstats.getsnapshot(stop)
         
local stats = jitstats.diffsnapshots(start, stop)
print(json.encode(stats))
```

javierguerragiraldez:
> The one thing i was thinking of adding was api to switch out the current stats state table.

one feature that would be very useful for us is to have a stats table per coroutine.  i think the vmevent hooks are executed in the same coroutine, unlike the profiler hook, right?

javierguerragiraldez:
(i just noticed that i had accidentally closed this, sorry for that)

jitstats.lua
```diff
@@ -464,9 +464,10 @@ function jitstats.print(statstbl, msg)
     end
 end
 
--- Turn off the jit for all the functions in this file so we don't contribute to the stats when being called. Sadly 
--- the effect of this call is deferred with respect to switching the function header bytecode to interpreter only 
--- mode to the first time each function trys to start a trace.
+-- Turn off the jit for all the functions in this file so they don't inflate the jit stats when called.
```
ltratt: "jit" -> "JIT".

jitstats.lua
```diff
@@ -464,9 +464,10 @@ function jitstats.print(statstbl, msg)
     end
 end
 
--- Turn off the jit for all the functions in this file so we don't contribute to the stats when being called. Sadly 
--- the effect of this call is deferred with respect to switching the function header bytecode to interpreter only 
--- mode to the first time each function trys to start a trace.
+-- Turn off the jit for all the functions in this file so they don't inflate the jit stats when called.
+-- Unfortunately these functions still decrement the hot counters until they reach zero. Instead of tracing
```
ltratt: "Instead of tracing" -> "Instead of tracing,"

jitstats.lua
```diff
@@ -464,9 +464,10 @@ function jitstats.print(statstbl, msg)
     end
 end
 
--- Turn off the jit for all the functions in this file so we don't contribute to the stats when being called. Sadly 
--- the effect of this call is deferred with respect to switching the function header bytecode to interpreter only 
--- mode to the first time each function trys to start a trace.
+-- Turn off the jit for all the functions in this file so they don't inflate the jit stats when called.
+-- Unfortunately these functions still decrement the hot counters until they reach zero. Instead of tracing
+-- the triggering bytecode is patched to interpreter only. This means no vm events are triggered for these points
```
ltratt: "vm" -> "VM"

ltratt:
OK, 3 minor typo things, and I think we're pretty much done.

ltratt:
I think this is ready for squashing.

fsfod:

ltratt:
Let's be consistent about whether commits end with "." or not (I prefer that they do end with a full stop, but I care more about being consistent). If you update, and do a force push, I should then be able to merge.

fsfod:
@ltratt ready now. I removed them from the two commits with them.

PR #13, opened 2017-10-06
Fix iterations_runner.lua to only load jitstats when instrumentation is requested

fsfod:
I some how missed this when i was doing quick test runs.


vext01:
Oops! Well spotted. Is this OK to merge as is?

fsfod:
yes

PR #14, opened 2017-10-18
Copy the build artifacts for each VM flavor to a sub directory

fsfod:
Instead of giving each vm builds executable a different name just copy to a sub folder along the with the builtin jit lua modules that will vary(mostly vmdef.lua) between build flavours this will be needed to fix the "require" error in krun.

### DIFF
--- a/luajit.krun
+++ b/luajit.krun
@@ -12,10 +12,13 @@ MAIL_TO = []
 
 DIR = os.getcwd()
 
-VM_ENV = {
-  'LUA_PATH' : '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(os.path.join(DIR, "lualibs")) + 
-               '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(os.path.join(DIR, "luajit_repo/src"))
-}
+VM_ENV = '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(os.path.join(DIR, "lualibs"))
+
+def make_env(vmdir):
+  basedir = os.path.join(DIR, "builds", vmdir)
+  return {
+    'LUA_PATH' : VM_ENV + '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(basedir) 
+  }
 
 HEAP_LIMIT = 2097152  # KiB
 STACK_LIMIT = 8192  # KiB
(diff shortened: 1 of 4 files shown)


ltratt:
This looks reasonable to me. OK @vext01?

ltratt:
Did you just rebase this?

luajit.krun
```diff
@@ -12,10 +12,10 @@ MAIL_TO = []
...
 DIR = os.getcwd()
 
-VM_ENV = {
-  'LUA_PATH' : '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(os.path.join(DIR, "lualibs")) + 
-               '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(os.path.join(DIR, "luajit_repo/src"))
-}
+VM_ENV = '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(os.path.join(DIR, "lualibs"))
+
+def make_env(vmdir):
+  return { 'LUA_PATH' : VM_ENV + '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(os.path.join(DIR, "builds/" + vmdir)) }
```
vext01: Can we split down this long line for ease of reading?

vext01: You are also mixing hard-coded `/` and `os.path.join` which feels a little odd.

Why don't you build the paths as lists, then `os.path.join` them at the end, before putting the semi colon between them. Would be clearer?

fsfod: I don't know what does and doesn't split across lines cleanly.

vext01: I mean, use intermediate assignments. But check the above comment about `ImportError` first.

fsfod:
> Can you do such a thing in Lua, and would it help?

The lua modules are what come with luajit itself if they can't be found for jitstats it should be an hard error.

vext01:
OK. I'll take your word for it (I don't pretend to fully understand).

So just that one thing about the hard to read lines.

May I suggest something like:

```python
paths = [
    ["blah", "blah", ...],
    ["blah", "blah", ...],
    ...
]
result = ";".join([os.path.join(x) for x in paths])
```

Untested, but you get the idea.

fsfod:
The base path is different for each list

vext01:
Well, you don't *have* to use string constants!

fsfod:
I think it just makes it overly complex for the number of paths joined.

vext01:
My version ends up with longer code, but (IMHO) more readable code. Up to you if you use it:

```python
def make_env(vm_subdir):
    lualibs = [DIR, "lualibs"]
    vm_dir = [DIR, "builds", vm_subdir]
    paths = [
        lualibs + ["?", "init.lua"],
        lualibs + ["?.lua"],
        lualibs + ["?", "?.lua"],
        vm_dir + ["?", "init.lua"],
        vm_dir + ["?.lua"],
        vm_dir + ["?", "?.lua"],
    ]
    print(paths)
    print(";".join([os.path.join(*x) for x in paths]))
```

(This code *is* tested).

It's actually possible to crunch that down further too, since you always append the same 3 paths to each base dir.

Like I say up to you...

fsfod:
i will just stick with the string formats for the current limited use case

vext01:
OK, do you want to squash?

fsfod:
@vext01 squashed and ready to merge

PR #15, opened 2017-11-10
Add few helper shell scripts 

fsfod:
resetkrun.sh clears out all the files generated from krun and quicktest.sh runs the bechmarks with a overlay krun file quicktest.krun that only runs one iteration for each benchmark

### DIFF
--- a/bencher.krun
+++ b/bencher.krun
@@ -1,0 +1,43 @@
+execfile("luajit.krun", globals()) # inherit settings from the main experiment
+
+MAIL_TO = ["fsfod11@gmail.com"]
+
+PRE_EXECUTION_CMDS = [
+    "sudo systemctl stop apt-daily.timer",
+    "sudo systemctl --runtime disable apt-daily.timer",
+
+    "sudo systemctl stop apt-daily-upgrade.timer",
+    "sudo systemctl --runtime disable apt-daily-upgrade.timer",
+
+    "sudo systemctl stop systemd-tmpfiles-clean.timer",
+    "sudo systemctl --runtime disable systemd-tmpfiles-clean.timer",
+
+    "sudo systemctl stop systemd-timesyncd",
+    "sudo systemctl --runtime disable systemd-timesyncd",
+
+    "sudo systemctl stop cron",
+]
+
+POST_EXECUTION_CMDS = [
+    "sudo systemctl start systemd-timesyncd || true",
+    "sudo systemctl start cron || true",
+
+    "sudo systemctl --runtime enable apt-daily.timer || true",
+    "sudo systemctl start apt-daily.timer || true",
+
+    "sudo systemctl --runtime enable apt-daily-upgrade.timer || true",
+    "sudo systemctl start apt-daily-upgrade.timer || true",
+
+    "sudo systemctl --runtime enable systemd-tmpfiles-clean.timer || true",
+    "sudo systemctl start systemd-tmpfiles-clean.timer || true",
+]
+
+def custom_dmesg_whitelist(defaults):
+    return defaults + [
+      "^.*en[0-9a-zA-Z]+: link (down|up)",
+      "^.*en[0-9a-zA-Z]+: Link is (up|down).*",
+      "^.*en[0-9a-zA-Z]+: Flow control is (on|off).*",
+      "^.*en[0-9a-zA-Z]+: EEE is (enabled|disabled).*",
+      "^.*random: crng init done",
+      "^.*apt-daily-upgrade.timer: Adding.*",
+    ]
--- a/resetkrun.sh
+++ b/resetkrun.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+function clean_results() {
+  (shopt -s nullglob; rm -rf $1*{log,manifest,bz2,envlogs,env,instr_data})
+}
+
+clean_results "luajit"
+clean_results "quicktest"
+clean_results "bencher"
--- a/.gitignore
+++ b/.gitignore
@@ -3,9 +3,11 @@ Thumbs.db
 
 # Visual Studio 2015 cache/options directory
 .vs/
-luajit_envlogs/
-luajit.manifest
-luajit_*.log
-luajit_results.json.bz2
-luajit_instr_data/
 /builds/
+*.manifest
+luajit_*.log
+quicktest_*.log
+*_envlogs/
+*_results.json.bz2
+*_instr_data/
+bencher_*.log
--- a/quicktest.krun
+++ b/quicktest.krun
@@ -1,0 +1,8 @@
+# Krun config file aimed to help quickly test all the benchmarks work under krun
+
+execfile("luajit.krun", globals()) # inherit settings from the main experiment
+
+for key, vm in VMS.iteritems():
+  vm["n_iterations"] = 2
+
+N_EXECUTIONS = 1
--- a/quicktest.sh
+++ b/quicktest.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+../krun/krun.py --no-pstate-check --no-tickless-check --quick --debug info --quick quicktest.krun


ltratt:
This is definitely one for Edd, so I'll leave him to it.

bencher.krun
```diff
@@ -0,0 +1,29 @@
+execfile("luajit.krun", globals()) # inherit settings from the main experiment
+
+MAIL_TO = ["fsfod11@gmail.com"]
```
vext01: Do you want your email address in the sources?

bencher.krun
```diff
@@ -0,0 +1,29 @@
+execfile("luajit.krun", globals()) # inherit settings from the main experiment
```
vext01: Did you mean to add this krun config file in this PR?

bencher.krun
```diff
@@ -0,0 +1,29 @@
...
+MAIL_TO = ["fsfod11@gmail.com"]
+
+PRE_EXECUTION_CMDS = [
+    "sudo systemctl stop apt-daily.timer",
+    "sudo systemctl --runtime disable apt-daily.timer", 
+    "sudo systemctl stop apt-daily-upgrade.timer",
+    "sudo systemctl --runtime disable apt-daily-upgrade.timer",
+    
+    "sudo systemctl stop systemd-timesyncd",
+    "sudo systemctl --runtime disable systemd-timesyncd",
```
vext01: Looks like you found a lot of stuff that I didn't. Will raise a PR to add these to the warmup experiment.

bencher.krun
```diff
@@ -0,0 +1,29 @@
...
+    "sudo systemctl stop cron",
```
vext01: Also `atd`

bencher.krun
```diff
@@ -0,0 +1,29 @@
...
+    "sudo systemctl start systemd-timesyncd || true",
+    "sudo systemctl start cron || true",
+]
+
+def custom_dmesg_whitelist(defaults):
+    return defaults + [
+      "^.*en[0-9a-zA-Z]+: link (down|up)",
+      "^.*en[0-9a-zA-Z]+: Link is (up|down).*",
+      "^.*en[0-9a-zA-Z]+: Flow control is (on|off).*",
+      "^.*en[0-9a-zA-Z]+: EEE is (enabled|disabled).*",
```
vext01: We should update the defaults in krun. The old defaults no longer work on debian 9 due to the new interface names in linux:

https://github.com/softdevteam/krun/blob/b022d26a3602b1e2d9fdea48648c7957707baf52/krun/platform.py#L1344-L1349

bencher.krun
```diff
@@ -0,0 +1,29 @@
...
+      "^.*random: crng init done",
+      "^.*apt-daily-upgrade.timer: Adding.*",
```
vext01: I'm surprise you see lines about the apt stuff you have disabled. This suggests they were not disabled properly. Similarly, RNG initialisation shouldn't happen mid-benchmark.

Did you hack `SYNC_WAIT` or supply a really small `TEMP_READ_PAUSE`?

resetkrun.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+(shopt -s nullglob; rm -rf luajit*{log,manifest,bz2,envlogs,env,instr_data})
+(shopt -s nullglob; rm -rf quicktest*{log,manifest,bz2,envlogs,env,instr_data})
```
vext01: Ah, `shopt -s nullglob`. This seems to be required since the update to Debian 9. Annoying.

bencher.krun
fsfod: krun will fail because systemd will write to stderr that the service doesn't exist.

bencher.krun
```diff
@@ -0,0 +1,29 @@
...
+    "sudo systemctl --runtime disable apt-daily-upgrade.timer",
+    
+    "sudo systemctl stop systemd-timesyncd",
+    "sudo systemctl --runtime disable systemd-timesyncd",
+    "sudo systemctl stop cron",
+]
+
+POST_EXECUTION_CMDS = [
+    "sudo systemctl start systemd-timesyncd || true",
+    "sudo systemctl start cron || true",
```
vext01: You need to bring up everything you turned off here.

bencher.krun
vext01: Did you want to take down the interfaces too?

bencher.krun
fsfod: well its just for the bencher machines this file

bencher.krun
fsfod: Well I didn't want to switch back and forth after I started iterating on it and it is a overlay file anyway.

bencher.krun
vext01: If `apt` is doing stuff during the benchmarks, then the results are compromised, so we would have to try and fix that.

bencher.krun
vext01: Oh right, maybe `atd` has been absorbed in to systemd, as seems to be the trend thesedays.

bencher.krun
fsfod: No the message is from systemd saying the timer entry had a random offset added to its trigger time

vext01: OK, now what about the RNG?

bencher.krun
fsfod: Maybe when its not on bencher3 i mean does krun even run the post commands if some pre commands failed. It just feels brittle with just a list of commands

bencher.krun
fsfod: Doe krun stop looking at dmesg when these are run?

bencher.krun
fsfod: Thats just that system has accumulated enough random noise for the rng as far as I understand.

bencher.krun
vext01: Yes, it always trys to run the post commands. We usually append `|| true` onto the non-essential POST commands, thus allowing them to fail.

bencher.krun
vext01: hrm, sounds fishy

bencher.krun
vext01: krun only looks at the dmesg during the process execution.

fsfod: well i was trying to figure out if that was the problem before

bencher.krun
vext01: Did you test this with `--quick`? I wonder if that is the culprit here...

vext01: Seems to me like the system isn't fully up before krun is running...

fsfod: The only option i used was hardware reboots passed to your cron script

fsfod:
@vext01 This is an example of the dmesg before:
```
New dmesg lines:
[    4.947994] r8169 0000:06:00.0 enp6s0: link up
  [    7.347645] systemd[1]: apt-daily.timer: Adding 8h 55min 56.499075s random time.
  [    7.347691] systemd[1]: apt-daily-upgrade.timer: Adding 39min 4.594621s random time.
  [    7.373053] systemd[1]: apt-daily-upgrade.timer: Adding 48min 44.386130s random time.
  [  286.007007] random: crng init done
```

vext01:
Something has gone wrong. Look at the timestamps on the dmesg lines you pasted. All but the RNG message happen in the first ten seconds, yet Krun is supposed to wait 3 minutes after reboot prior to executing a pexec. Are you sure you didn't use `--quick` by accident?

fsfod:
I don't know I just picked one out of my emails that had the lines.

vext01:
I'm going to need you to look into it I'm afraid. I can't do much more without the full details.

fsfod:
@vext01  I tested it with only hardware reboots and the rng messages still happen and that timestamp is not at 10 seconds

fsfod:
Another recent example
```
[   39.074516] systemd[1]: apt-daily-upgrade.timer: Adding 28min 54.462518s random time.
[  167.179970] random: crng init done
```

fsfod:
Debug log snippet :

```
[2017-11-15 20:19:22: INFO] stderr: [iterations_runner.lua] iteration 1500/1500
[2017-11-15 20:19:22: INFO] Finished 'series(2)' (default-lua variant) under 'RaptorJIT'
[2017-11-15 20:19:22: DEBUG] execute shell cmd: /usr/bin/sudo -u krun rm /tmp/envlog-PTJxcD.env
[2017-11-15 20:19:22: DEBUG] Update manifest flag: C
[2017-11-15 20:19:22: DEBUG] Reading status cookie from /home/tfransham/work/lua_benchmarking/bencher.manifest
[2017-11-15 20:19:22: DEBUG] Reading status cookie from /home/tfransham/work/lua_benchmarking/bencher.manifest
[2017-11-15 20:19:22: DEBUG] execute shell cmd: dmesg
[2017-11-15 20:19:22: DEBUG] Allowed dmesg change: '[    4.937955] r8169 0000:06:00.0 enp6s0: link up'
[2017-11-15 20:19:22: DEBUG] Allowed dmesg change: '[    4.937960] IPv6: ADDRCONF(NETDEV_CHANGE): enp6s0: link becomes ready'
[2017-11-15 20:19:22: WARNING] New dmesg lines:
[    7.214546] systemd[1]: apt-daily-upgrade.timer: Adding 40min 58.534506s random time.
[  208.717416] random: crng init done
```

vext01:
Looks to me like Krun is collecting the dmesg too early.

vext01:
Leave it with me.

vext01:
OK Tom, can you try with the latest Krun and put the full debug log somewhere. Then we can look at which dmesg lines we really need to allow.

A handful of pexecs will be fine.

fsfod:
@vext01  Well it looks like what i assumed from seeing systemd debug reports about the message is trigged when you stop service or the system time is updated.
```
New dmesg lines:
[  123.379282] systemd[1]: apt-daily-upgrade.timer: Adding 55min 36.727794s random time.
```

vext01:
OK, so we don't see any early dmesg lines showing up now?

fsfod:
Yeah only after 120 seconds 
```
[  123.379282] systemd[1]: apt-daily-upgrade.timer: Adding 55min 36.727794s random time.
[  186.470624] random: crng init done
```

```
[  126.604025] systemd[1]: apt-daily-upgrade.timer: Adding 44min 54.570164s random time.
[  147.425788] random: crng init done
```

vext01:
OK. So I think you just need to bring up in POST_EXEC what you took down in PRE_EXEC, then we are done here.

vext01:
OK, but I think we are playing with fire here.

I've noted that bringing dbus back up again doesn't bring up the dependants, so I will need to explicitly kill and start those too.

vext01:
(sorry, posted on wrong PR, deleted)

vext01:
Is there a reason why some of the service commands use `enable/disable` and others use `start/stop`. I presume you want the latter, so the changes are temporary?

fsfod:
I only used stop for timers in pre commands in case it was already running

vext01:
Should we be consistent?

fsfod:
Well comparing the output from systemctl list-timers it looks like you still manually still start disabled timers and for a timer to be considered inactive it has to be disabled and stopped.

vext01:
ok

vext01:
are you ready to squash?

fsfod:
yes

vext01:
go ahead

fsfod:
@vext01 ready to merge

PR #16, opened 2017-11-15
Add lua-vermelha as a optional Lua runtime

fsfod:
Mostly just builds the binary and copies it to the builds folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ quicktest_*.log
 *_results.json.bz2
 *_instr_data/
 bencher_*.log
+lua-vermelha/
+raptorjit_repo/
+luajit_repo/
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: c
 sudo: false
 
 before_script:
-  - ./build_luajit.sh
+  - ./build.sh
 
 script:
   - ./builds/nojit/luajit simplerunner.lua
--- a/build.sh
+++ b/build.sh
@@ -2,32 +2,32 @@
 
 #Build the various interesting flavours of LuaJIT
 if [ $# -gt 1 ]; then
-    echo "Usage: build.sh [<LuaJIT source dir>]" 2>&1
-    exit 1
+  echo "Usage: build.sh [<LuaJIT source dir>]" 2>&1
+  exit 1
 fi
 
 if [ $# -eq 1 ]; then
-    ljsrc=$1/src
-    echo "Using ${ljsrc} as LuaJIT source path"
+  ljsrc=$1/src
+  echo "Using ${ljsrc} as LuaJIT source path"
 else
-    echo "Defaulting to ./luajit_repo/src as LuaJIT source path"
-    ljsrc=luajit_repo/src
+  echo "Defaulting to ./luajit_repo/src as LuaJIT source path"
+  ljsrc=luajit_repo/src
 
-    if [ ! -d "luajit_repo" ]; then 
-      git clone https://github.com/softdevteam/LuaJIT luajit_repo
-    fi
+  if [ ! -d "luajit_repo" ]; then
+    git clone https://github.com/softdevteam/LuaJIT luajit_repo
+  fi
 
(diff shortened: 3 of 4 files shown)


build_luav.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+#Build the various intresting flavors of LuaJIT
```
ltratt: Typo "interesting".

ltratt:
I suggest we add this to `build_luajit.sh` and rename that (in this or another PR) to `build.sh`.

fsfod:
I was just making it optional as a separate file

ltratt:
I suggest that we make it non-optional as part of the build (it is quick to build, after all). Otherwise we'll end up with loads of `build_` scripts that we'll get confused by.

fsfod:
Well travis runs that as well and it can't even build it even though it wasted over 10 mins trying

ltratt:
I'm fine with guarding the build of this VM on Travis. You can check an environment variable for that IIRC.

fsfod:
Well i should add it crashes the compiler on my Linux machine and on my laptop it fails to build with make -j only non parallel build works.

ltratt:
Where are we with this?

fsfod:
Was there a better branch we should be using i think you said before?

ltratt:
If we guard the build, I don't think it matters?

fsfod:
It was guarded for travis atleast. idk what option you want it under maybe BUILDALL

ltratt:
Something like that sounds plausible.

fsfod:
@ltratt I added a guard for it and and theres now main ./build.sh file setup in a previous commit

ltratt:
Thanks. What does "theres now main ./build.sh file setup in a previous commit" mean though?

fsfod:
i removed build_luav.sh and just added its contents into a new ./build.sh file that also builds luajit

ltratt:
Ah, I see. I think what I was suggesting was renaming `build_luajit.sh` to `build.sh` and then putting the luav stuff in that new `build.sh`. Can we do something along those lines please?

fsfod:
i wanted to keep build_luajit.sh seperate

ltratt:
If we do that, we're going to end up with loads of `build_` files which are going to be confusing. It's not like a single `build.sh` is going to be huge.

fsfod:
There won't be anymore any new ones will go in to build.sh

ltratt:
OK, then merging them is even easier. [I'm always open to reconsidering my position, but "I want" isn't an adequate reason.]

fsfod:
It will just be the main file i use for development i just want just something simple to run that i don't have remember to use any special command line.

ltratt:
What special command line is needed for `build.sh`?

fsfod:
none yet because you said to guard luav

ltratt:
Perfect. Then there's nothing stopping merging the two files. If you do that, we should be able to merge this soon. Thanks!

fsfod:
well i guess build_luajit was designed to optionally take a path to custom luajit dir

ltratt:
It can still do that when it's merged in.

fsfod:
@ltratt well there combined now

ltratt:
Thanks!

One style comment: the indentation of `build.sh` is all over the place in parts e.g.:

```sh
+else
+    echo "Defaulting to ./luajit_repo/src as LuaJIT source path"
+    ljsrc=luajit_repo/src
+
+    if [ ! -d "luajit_repo" ]; then 
+      git clone https://github.com/softdevteam/LuaJIT luajit_repo
+    fi
```

Can you add a separate commit which fixes this?

fsfod:
@ltratt fixed

ltratt:
Thanks. If you can squash this down (please leave the indentation commit separate), then I can merge.

fsfod:
@ltratt why do you want to keep the indentation commit the file always has lots of churn before that

ltratt:
To reduce the churn.

fsfod:
@ltratt well the code is squashed

PR #18, opened 2017-12-04
Fix raptorjit crashing

fsfod:
Start using a commitlock to checkout a known good raptorjit commit


fsfod:
@ltratt this is needed for the benchmarks to run cleanly on bencher9

PR #20, opened 2017-12-18
Add support for building all the LuaJIT flavors with extra XCFLAGS 

fsfod:
Allow building all the LuaJIT flavours with extra XCFLAGS in build_luajit.sh allowing us to easily define HOTCOUNT_SIZE_SCALE.

### DIFF
--- a/build_luajit.sh
+++ b/build_luajit.sh
@@ -42,25 +42,27 @@ function copy_binaries() {
     cp ${ljsrc}/libluajit.so ${ljbins}/$1/libluajit.so
     cp ${ljsrc}/jit/*.lua ${ljbins}/$1/jit/
 }
-  
+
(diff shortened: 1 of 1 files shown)


ltratt:
I'm not sure we need this PR. If the user just does `XCFLAGS=blah make` it'll get passed through to the LuaJIT builds, right?

fsfod:
XCFLAGS is already being set its what is controlling the build flavour.

ltratt:
Set where? I don't understand.

fsfod:
in build_luajit.sh for [every build flavour except the first](https://github.com/softdevteam/lua_benchmarking/pull/20/files#diff-7359cef4a35217e8bac62caca63f7361L53)

ltratt:
Ah :/ Hmm, OK, I guess it's difficult to do this neatly then.

PR #22, opened 2018-01-11
A first attempt at expanding the README.

ltratt:
Please check this carefully -- I may well have got details wrong! Suggestions for additions and  improvements are also welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,29 @@
-# lua_benchmarking
+# Benchmarking Lua
+
+This repository collects together a number of open-source Lua benchmarks,
+suitable for quick or rigorous benchmarking. Contributions are welcome!
+
+
+## Adding new benchmarks
+
+New benchmarks should be put in the `benchmarks/` repository with an appropriate
+name. The "main" file should be called `bench.lua`, which must contain a
+function `run_iter` which takes a single parameter `n` which is the number of
+times the benchmark should be run in a `for` loop (or equivalent) to make up a
+single "in-process iteration". The reason for this is that many benchmarks run
+too quickly for reliable measurements to be taken. However, the number of times
+a benchmark should be repeated to run "long enough" is machine dependent.
+`run_iter` is thus easily customisable for different situations. Roughly
+speaking, a single in-process iteration should run for around 1 second (with a
+minimum acceptable of 0.1s).
+
+
+## Benchmarking using Krun
+
+Those who wish to use [Krun](http://soft-dev.org/src/krun/) or
+[warmup_stats](http://soft-dev.org/src/warmup_stats/) should investigate the
+various `.krun` files in this repository. `quicktest.krun` is useful to
+determine whether the benchmark suite and benchmarking setup are free
+from errors, but does not run benchmarks long enough to give useful statistics.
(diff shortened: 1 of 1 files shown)


fsfod:
quicktest.krun is really more a sanity check it only runs the benchmarks once for one iteration and execution

ltratt:
Point taken. See if that new commit helps.

fsfod:
Thats a better explanation. There should also be some note about Lua modules living in the lualibs folder

ltratt:
I don't know what those are, so you're going to have to explain them to me...

fsfod:
They have to be in that folder because require/module loading is not relative to a script in Lua

ltratt:
Can you propose some text and tell me where you think it should go? I don't really know this well enough.

fsfod:
something like Any extra Lua modules your benchmark needs should go in the 'lualibs' folder

ltratt:
I need more than that. Why would there be extra Lua modules? etc. etc.

fsfod:
because the benchmark might be based on them?

ltratt:
Please can you draft a paragraph that's clear about what you're thinking. I'm not able to guess well enough to do a proper job :/

fsfod:
"If you benchmark depends on any custom Lua modules should put them in the 'lualibs' folder so they are found correctly by 'require'"

ltratt:
Issue https://github.com/softdevteam/lua_benchmarking/issues/24 means that we can more-or-less ignore the shared global dir at the moment. I think that means this PR is ready for squashing/

ltratt:
Ping.

fsfod:
Maybe there should be some note about about using the ffi/ctype system since we do use it it some of the benchmarks

ltratt:
Happy to consider it -- you'll need to suggest draft text for that though.

fsfod:
something like "Some of the existing benchmarks already use LuaJIT's ffi/cdata system so feel free to use it in your own benchmarks if it makes sense"

ltratt:
Hmm, maybe the real thing here is that some benchmarks might need to build libraries (etc.) in order to run. Maybe we should do a (separate) PR that says "benchmarks can have a `build.lua` script which will be run before benchmarking in order to build XYZ"? That would need a bit of thinking about to get the design right, but it might not be the worst idea ever.

fsfod:
I meant cdata system thats part of ffi system but doesn't involve calling out to external code

ltratt:
cdata?

fsfod:
it extra type in LuaJIT that mimics stuff from C like native structs and pointers

ltratt:
I'm sorry, you've completely lost me. How does that relate to the sentence you wrote?

fsfod:
It might be more work than just using shell scripts. Lua does have a module repository system like pip like system called LuaRocks

ltratt:
The code can just call a shell script if it wants. Or we could call `build.sh` in the directory.

OK, I'm going to suggest that I raise an issue to debate this point further. But I think this PR is now ready to be squashed.

fsfod:
@ltratt squash away

ltratt:
Thanks! Squashed.

Issue #23, opened 2018-01-11
A simple way of quickly running benchmarks


ltratt:
I think we need to provide a simple way of people who don't have Krun to run benchmarks quickly, using whatever Lua VM they have knocking around, to encourage people to contribute to this repo. At its most basic it should just be something like:

```sh
$ lua quick_run.lua
Running bench1: .........................
  Mean: 2.2s +/- 0.1s
Running bench2: .........................
  Mean: 0.8s +/- 0.05s
```

What we need to do is a get a reasonable trade-off between ease of use and not being completely statistically misleading. My suggestion is that we run each benchmark for 30 in-process iterations (printing a "." to stdout as each is being run, so the user knows something is happening), and after each printing a mean and a 95% confidence interval. `quick_run.lua` could take an optional integer parameter which varies the number of in-process iterations (so `quick_run.lua 10` would run 10 in-process iterations).

We would of course need to put a warning in the README along the lines of "this isn't statistically reliable, but it is better than nothing." I can handle that part if you want me to.

fsfod:
simplerunner.lua does about half this already

ltratt:
Aha, that's a good start :) Let's get it doing the rest outlined above and then we're sorted!

fsfod:
> whatever Lua VM they have knocking around
I think it will be pretty limiting to depend on plain Lua and not some build of LuaJIT with the ffi system i can callout in to the os and collect better data.

ltratt:
Can one not install the FFI with plain Lua?

fsfod:
That would be a binary module i was trying avoid to having to include any

ltratt:
Is building it difficult? If so, we could presumably say "if you want to run this on plain Lua you will need to have installed XXX yourself"?

ltratt:
#26 mostly addressed this, but didn't add any documentation to the README. @tfransham can you add a few lines to the README that tell people how to run things?

ltratt:
I've started looking into fixing this, so the action is now on me.

ltratt:
Fixed in #28.

PR #26, opened 2018-01-16
Revamp simplerunner.lua to make running/creating benchmarks easier 

fsfod:
This is revamp of simplerunner.lua to help achieve the goal of #23 for providing a way to quickly run benchmarks and help in the process of making them.
There is now a basic command line option system borrowed from dynasm.lua. Some basic stats are now calculated and printed from all the iteration times. 
Passing no command line augments to simplerunner will cause it to run all the benchmarks each for 30 iterations and a default scaling of one. To run a single benchmark just pass benchmark name as the last command line entry. An optional iteration count can be specified after it like so

`simplerunner.lua luafun 1500`

Added basic progress indicator by printing a dot to the console for each iteration run.
Made collecting and printing jitstats off by default now..
Turned off JITing benchmark loop to try to keep the timing more stable.
Added optional support for using a more high res timer by using the ffi system otherwise os.clock is used. Also the functions used to get the highres time are also pre JIT'ed to try keep the timing stable.

### DIFF
--- a/lualibs/time.lua
+++ b/lualibs/time.lua
@@ -1,0 +1,152 @@
+
+--wall-clock time, monotonic time and sleeping for Windows, Linux and OSX.
+--Written by Cosmin Apreutesei. Public Domain.
+
+local ffi = require'ffi'
+
+local M = {}
+local C = ffi.C
+
+if ffi.os == 'Windows' then
+
+	ffi.cdef[[
+	void time_GetSystemTimeAsFileTime(uint64_t*) asm("GetSystemTimeAsFileTime");
+	int  time_QueryPerformanceCounter(int64_t*) asm("QueryPerformanceCounter");
+	int  time_QueryPerformanceFrequency(int64_t*) asm("QueryPerformanceFrequency");
+	void time_Sleep(uint32_t ms) asm("Sleep");
+	]]
+
+	local t = ffi.new'uint64_t[1]'
+	local DELTA_EPOCH_IN_100NS = 116444736000000000ULL
+
+	function M.time()
+		C.time_GetSystemTimeAsFileTime(t)
+		return tonumber(t[0] - DELTA_EPOCH_IN_100NS) / 10^7
+	end
+
+	assert(C.time_QueryPerformanceFrequency(t) ~= 0)
+	local qpf = tonumber(t[0])
+
+	function M.clock()
+		assert(C.time_QueryPerformanceCounter(t) ~= 0)
+		return tonumber(t[0]) / qpf
+	end
+
+	function M.sleep(s)
+		C.time_Sleep(s * 1000)
+	end
+
+elseif ffi.os == 'Linux' or ffi.os == 'OSX' then
+
+	ffi.cdef[[
+	typedef struct {
+		long s;
+		long ns;
+	} time_timespec;
+
+	int time_nanosleep(time_timespec*, time_timespec *) asm("nanosleep");
+	]]
+
+	local EINTR = 4
+
+	local t = ffi.new'time_timespec'
+
+	function M.sleep(s)
+		local int, frac = math.modf(s)
+		t.s = int
+		t.ns = frac * 10^9
+		local ret = C.time_nanosleep(t, t)
+		while ret == -1 and ffi.errno() == EINTR do --interrupted
+			ret = C.time_nanosleep(t, t)
+		end
+		assert(ret == 0)
+	end
+
+	if ffi.os == 'Linux' then
+
+		ffi.cdef[[
+		int time_clock_gettime(int clock_id, time_timespec *tp) asm("clock_gettime");
+		]]
+
+		local CLOCK_REALTIME = 0
+		local CLOCK_MONOTONIC = 1
+
+		local clock_gettime = ffi.load'rt'.time_clock_gettime
+
+		local function tos(t)
+			return tonumber(t.s) + tonumber(t.ns) / 10^9
+		end
+
+		function M.time()
+			assert(clock_gettime(CLOCK_REALTIME, t) == 0)
+			return (tos(t))
+		end
+
+		function M.clock()
+			assert(clock_gettime(CLOCK_MONOTONIC, t) == 0)
+			return (tos(t))
+		end
+
+	elseif ffi.os == 'OSX' then
+
+		ffi.cdef[[
+		typedef struct {
+			long    s;
+			int32_t us;
+		} time_timeval;
+
+		typedef struct {
+			uint32_t numer;
+			uint32_t denom;
+		} time_mach_timebase_info_data_t;
+
+		int      time_gettimeofday(time_timeval*, void*) asm("gettimeofday");
+		int      time_mach_timebase_info(time_mach_timebase_info_data_t* info) asm("mach_timebase_info");
+		uint64_t time_mach_absolute_time(void) asm("mach_absolute_time");
+		]]
+
+		local t = ffi.new'time_timeval'
+
+		function M.time()
+			assert(C.time_gettimeofday(t, nil) == 0)
+			return tonumber(t.s) + tonumber(t.us) / 10^6
+		end
+
+		--NOTE: this appears to be pointless on Intel Macs. The timebase fraction
+		--is always 1/1 and mach_absolute_time() does dynamic scaling internally.
+		local timebase = ffi.new'time_mach_timebase_info_data_t'
+		assert(C.time_mach_timebase_info(timebase) == 0)
+		local scale = tonumber(timebase.numer) / tonumber(timebase.denom) / 10^9
+		function M.clock()
+			return tonumber(C.time_mach_absolute_time()) * scale
+		end
+
+	end --OSX
+
+end --Linux or OSX
+
+if not ... then
+	io.stdout:setvbuf'no'
+	local time = M
+
+	print('time ', time.time())
+	print('clock', time.clock())
+
+	local function test_sleep(s, ss)
+		local t0 = time.clock()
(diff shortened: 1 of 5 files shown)


lualibs/time.lua
```diff
@@ -0,0 +1,152 @@
...
+--wall-clock time, monotonic time and sleeping for Windows, Linux and OSX.
+--Written by Cosmin Apreutesei. Public Domain.
+
+local ffi = require'ffi'
+
+local M = {}
+local C = ffi.C
+
+if ffi.os == 'Windows' then
+
```
ltratt: Blank line... and this file uses tabs for indentation?

simplerunner.lua
```diff
@@ -31,30 +56,51 @@ local benchlist = {
...
+if timeloaded then
+    clock = time.clock
+ 
+    if jit_util then
+        -- Force the timer functions to get JIT'ed so they don't go through the slow ffi path in the interpreter thats data driven
+        jitfunc(startimer)
+        jitfunc(stoptimer)
     end
+else
+    -- Uses os.clock uses the libc 'clock' function. CLOCKS_PER_SEC is normally 1Mhz on Linux but on windows its was 1khz.
```
ltratt: "Uses" twice?

simplerunner.lua
```diff
@@ -63,53 +109,276 @@ function runbench(name, count)
...
-        local start = os.clock()
-        run_iter(count)
-        print(name.." took", os.clock() - start)
+       times = {}
     end
+    io.write("Running "..name..": ")
+    io.flush()
 
+    local run_iter = _G.run_iter
+    for i= 1, count do
```
ltratt: Missing space before `=`?

simplerunner.lua
```diff
@@ -63,53 +109,276 @@ function runbench(name, count)
...
+    
+    if loading_jitstats and jitstats.stats.starts > 0 then 
+        jitstats.print("Traces created while loading benchmark")
+        jitstats.reset()
+    end  
+end  
 
-    count = count or 1
+function runbench(name, count, scaling)
+    scaling = scaling or 1
```
ltratt: I'm not sure I see the point of this line.

simplerunner.lua
```diff
@@ -63,53 +109,276 @@ function runbench(name, count)
...
+    loadbench(name)
```
ltratt: Can we unload the benchmark at the end of the benchmarking?

simplerunner.lua
```diff
@@ -63,53 +109,276 @@ function runbench(name, count)
...
+
+-- Short aliases for long options.
+local opt_alias = {
+    h = "help", ["?"] = "help", V = "version",
+    b = "bench", c = "count", s = "scaling",
+}
+
+-- Print help text.
+function opt_map.help()
+  io.stdout:write[[
```
ltratt: Two space indent?

simplerunner.lua
```diff
@@ -63,53 +109,276 @@ function runbench(name, count)
...
+-- Misc. options.
+function opt_map.count(args) g_opt.count = optparam(args, "number") end
+function opt_map.bench(args) g_opt.bench = optparam(args) end
+function opt_map.scaling(args) g_opt.scaling = optparam(args, "number") end
+function opt_map.jitstats() g_opt.jitstats = true end
+function opt_map.benchloadjitstats() loading_jitstats = true end
+function opt_map.nogc() g_opt.nogc = true end
+function opt_map.jitlog() g_opt.jitlog = true end
+
+------------------------------------------------------------------------------
```
ltratt: Not sure this line is helpful.

ltratt:
Does Lua not have a command-line parameter parsing library? Writing our own feels a bit retrograde!

simplerunner.lua
```diff
@@ -342,25 +342,22 @@ local function parseargs(args)
...
     local nargs = #args - args.argn + 1
-    if not g_opt.bench and g_opt.count then
-        if nargs == 0 then
-            -- Show help and exit
-            opt_map.help()
+    if (not g_opt.bench or not g_opt.count) and nargs > 0 then
+        local arg1 = args[args.argn]
+        local num = tonumber(arg1)
+        if num then
+            -- Just a iteration count was passed ignore any extra args after
```
ltratt: "aN iteration count". Missing ";" before "ignore"?

simplerunner.lua
```diff
@@ -342,25 +342,22 @@ local function parseargs(args)
...
-            else
-                local arg2 = args[args.argn + 1]
-                g_opt.bench = arg1
-                if arg2 then
-                    num = tonumber(arg2)
-                    if not num then
-                        opterror("Bad count value '", arg2, "' to run benchmarks for")
-                    end
-                    g_opt.count = num
+            -- We expect a benchmark name followed by a optional iteration count
```
ltratt: "aN" optional.

lualibs/time.lua
fsfod: This a Lua external module that i didn't make

ltratt: Why are we copying it into our repo and not downloading it in a build script? We don't want to maintain forks of such things if we can avoid it...

fsfod: Its not a Lua rocks module https://github.com/luapower/time so there is no a package manager to get it from.

simplerunner.lua
fsfod: well i do nil out the run_iter global function but that doesn't really do much unless we run the garbage collector

lualibs/time.lua
ltratt: We could just download it via this link `https://raw.githubusercontent.com/luapower/time/master/time.lua`.

simplerunner.lua
ltratt: I think it would be fine to run the garbage collector after an unload. Running multiple benchmarks in the same process has weird effects; anything we can do to lessen those effects is a good idea (even though I know we can't do *that* much).

ltratt:
I have a simple feature request to stick in this PR: let's randomise the order of benchmarks whenever we run them (even if the user specifies benchmarks, we'll randomise that order). That way, if you run the benchmark suite twice, you're less likely to be caught out by the effects that one benchmark might have on the next. [This is an imperfect mitigation strategy, but it's better than nothing.]

vext01:
Good idea.

lualibs/time.lua
ltratt: I think everyone can get hold of `curl`/`wget` and we should assume that someone capable of using this repository knows how to do so. Duplicating code is just a recipe for disaster long-term: it means we don't get bug fixes etc.

fsfod:
> Does Lua not have a command-line parameter parsing library? Writing our own feels a bit retrograde!
Lua only has a minimal standard library

> Can we run things on Travis for less time to save them some energy. I'd have thought a parameter of 2 or 3 would be sufficient.
I already reduced the time benchmarks run for travis. Most of the time is spent building all the flavours of Lua. We also don't have any scaling values setup in this runner

simplerunner.lua
fsfod: Its defaulting it to 1 or if the parameter is unset

ltratt: I can't see anywhere where the parameter is unset though?

simplerunner.lua
fsfod: its was there from the code i copied it from although the ordering is slightly different

ltratt: Copied from where? What's the license of the thing that was copied?

simplerunner.lua
fsfod: well the run_all_benchmarks call at the bottom doesn't pass it and i think passing a scaling value to run_all_benchmarks would be an override from a preconfigured scaling settings system we don't have

ltratt: I'm not sure I understand the sentence after "i think passing" -- can you rephrase?

fsfod: I think we should have our benchmark scaling values in a shared json file used by both our krun file and this

ltratt: I'm fine with that, at which point we wouldn't need this default?

fsfod:
@ltratt I've added benchmark randomization, garbage collection after each benchmark and per benchmark scaling from a json config file.

simplerunner.lua
```diff
@@ -95,7 +95,7 @@ if timeloaded then
         jitfunc(stoptimer)
     end
 else
-    -- Uses os.clock uses the libc 'clock' function. CLOCKS_PER_SEC is normally 1Mhz on Linux but on windows its was 1khz.
+    -- Fallback to os.clock this uses the libc 'clock' function. CLOCKS_PER_SEC is normally 1Mhz on Linux but on windows its was 1khz.
```
ltratt: "Fallback to os.clock WHICH uses".

simplerunner.lua
```diff
@@ -167,10 +167,25 @@ if jit then
     jit.off(runbench)
 end
 
-function run_all_benchmarks(count, scaling)
-    assert(#benchlist ~= 0)
+function permute(tab)
+    local count = #tab
+    math.randomseed(os.time())
+    for i, v in ipairs(tab) do
+        local j = math.random(i, count)
```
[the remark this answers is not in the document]
ltratt: I don't think it's quite equivalent: look at the upper bound, for example. [I think you've almost, but not quite, implemented the version Wikipedia describes as "An equivalent version which shuffles the array in the opposite direction (from lowest index to highest) is:"]

fsfod:
@ltratt i fixed the shuffle and various little things

ltratt:
Thanks. If you squash, I can merge.

fsfod:

PR #28, opened 2018-01-25
Simplerunner instructions

ltratt:
Make people aware of `simplerunner.lua`'s existence and how they can run it.

While here, fix `simplerunner.lua`'s description of its arguments, which had an error in it, and was also a bit misleading.

### DIFF
--- a/simplerunner.lua
+++ b/simplerunner.lua
@@ -309,7 +309,7 @@ local opt_alias = {
 function opt_map.help()
     io.stdout:write[[
 
-Usage: simplerunner [OPTION]...  benchmark, count
+Usage: simplerunner [OPTION] [<benchmark>] [<count>]
 
   -h, --help           Display this help text.
   -b, --bench name     Name of benchmark to run.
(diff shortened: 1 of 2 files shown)


simplerunner.lua
```diff
@@ -309,7 +309,7 @@ local opt_alias = {
 function opt_map.help()
     io.stdout:write[[
 
-Usage: simplerunner [OPTION]...  benchmark, count
+Usage: simplerunner [OPTIONS] [<benchmark_1> ... <benchmark_n>] [count]
```
fsfod: The trailing argument parsing was very simplistic it only accepts one benchmark name

ltratt: Ah yes, you're right. I'll fix the docs.

ltratt:
@tfransham See if that new commit works for you.

fsfod:
Its fine maybe we could say use multiple -b options to run your own custom list of benchmarks

ltratt:
No, I don't think that's a good approach. https://github.com/softdevteam/lua_benchmarking/issues/30 outlines a better approach IMHO.

Can I squash this PR?

ltratt:
@tfransham Ping.

fsfod:
yes

ltratt:
Squashed and ready for merging.

PR #31, opened 2018-01-25
Add support for specifying more than once benchmark in the trailing args

fsfod:
Also added stricter checks for the optional count arg at the end of the list is a number and not another benchmark name

### DIFF
--- a/simplerunner.lua
+++ b/simplerunner.lua
@@ -386,25 +386,11 @@ local function parseargs(args)
         load_jitstats()
     end
 
-    -- Check for proper number of arguments.
+    -- Check for a trailing list of benchmark names.
     local nargs = #args - args.argn + 1
-    if (not g_opt.bench or not g_opt.count) and nargs > 0 then
-        local arg1 = args[args.argn]
-        local num = tonumber(arg1)
-        if num then
-            -- Just a iteration count was passed; ignore any extra args after.
-            g_opt.count = num
-        else
-            -- We expect a benchmark name followed by an optional iteration count.
-            local arg2 = args[args.argn + 1]
-            table.insert(g_opt.benchmarks, arg1)
-            if arg2 then
-                num = tonumber(arg2)
-                if not num then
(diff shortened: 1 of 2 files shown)


ltratt:
I like part of this, but we need to get rid of the "if the last thing looks like a number make it a count". As discussed in #31 it's a gun pointed at the foot of users. It's not worth the danger merely to save people typing "-c".

fsfod:
It was not the extra typing -c its more it will be in the middle of the command line making to slower to quickly edit and run again

ltratt:
I'm sorry, I have no idea what that means.

fsfod:
The count will not be at the end of the command line in the terminal when you want to rerun it and quickly add/remove a 0 from it

ltratt:
Sorry, I don't buy that as a sufficient reason to leave a gun pointed at people's feet. There are various ways to edit a previous command easily in the shell. Even `Ctrl+A` largely solves the problem (but one can get more clever than that if it's *really* necessary).

Let's please remove the "count as extra argument" feature entirely.

fsfod:
it was also used as the only argument

ltratt:
Thanks for the new commit. I think this can be squashed ready for merging.

fsfod:

simplerunner.lua
```diff
@@ -386,25 +386,11 @@ local function parseargs(args)
         load_jitstats()
     end
 
-    -- Check for proper number of arguments.
+    -- Check for list of benchmark names followed by an optional iteration count.
```
ltratt: I think this comment is now inaccurate?

ltratt:
@tfransham ping

fsfod:
@ltratt  i left the commit for you to check before squashing it again or you wouldn't know what i changed

ltratt:
You didn't tell me if you were finished or not, so I wasn't sure what the status was.

This looks fine now. I think we can now squash this ready for merging.

fsfod:

PR #33, opened 2018-01-30
Add benchmark metadata system

fsfod:
Add Benchmark metadata system stored in the same file as benchmark scaling values benchinfo.json and use it to fix running under normal Lua by skipping benchmarks flagged as using the ffi system. 
The list benchmark names from the scaling table in benchinfo.json is now used the base list of benchmarks to run instead of a hard coded table in simplerunner.lua

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,14 +27,40 @@ better quality statistics) -- see `simplerunner.lua -h` for more information.
 
 New benchmarks should be put in the `benchmarks/` repository with an appropriate
 name. The "main" file should be called `bench.lua`, which must contain a
-function `run_iter` which takes a single parameter `n` which is the number of
+function `run_iter` which takes a scaling parameter `n` which is the number of
 times the benchmark should be run in a `for` loop (or equivalent) to make up a
 single "in-process iteration". The reason for this is that many benchmarks run
 too quickly for reliable measurements to be taken. However, the number of times
 a benchmark should be repeated to run "long enough" is machine dependent.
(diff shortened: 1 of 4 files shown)


ltratt:
Can the FFI not be built at all on normal Lua?

fsfod:
People have made binary Lua modules to try and replicate it for normal Lua

fsfod:
Well the facebook one errors on capnproto so i guess incomplete/limitations of being a separate module instead of a integrated system.

fsfod:
the only entry i added so far is 'ffi' boolean maybe should be called something like ffirequired. Could maybe have build commands in there

ltratt:
`ffirequired` seems like a good idea. How would you suggest going about putting in build commands?

fsfod:
I was thinking like a list of shell commands to run in the benchmark's directory or something

ltratt:
On reflection, I think it might just be easier to have an optional `build.sh` file in each benchmark directory: if it's there the runner runs it. The only tricky bit is should the runner or the build file decide if things have already been built?

fsfod:
I was thinking maybe benchmark metadata should specify some file or files that should exist otherwise run the build or just have some config file that we record that we built it

ltratt:
I now think it's overkill. If `build.sh` exists, then there is something that could be built; and the `build.sh` file itself is best placed to work out if it needs to rebuild anything.

ltratt:
Looks good. Time to squash!

fsfod:
ready for merge

PR #34, opened 2018-02-01
Fix bad 'count' value in the stats calculation

fsfod:
The count value was off by one causing some of the stats to go to infinity with low benchmark run counts.

### DIFF
--- a/simplerunner.lua
+++ b/simplerunner.lua
@@ -241,13 +241,12 @@ function run_benchmark_list(benchmarks, count, options)
     end
 end
 
-function calculate_stats(times, start)
+function calculate_stats(times)
     local stats = {}
-    start = start or 1
-    local count = #times - start-1
+    local count = #times
 
     local min, max, total = 10000, 0, 0
-    for i = start, #times do
+    for i = 1, count do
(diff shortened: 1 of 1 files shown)


ltratt:
This looks reasonable. Can you add some explanation to the commit message please? Then I think we'll be ready to merge.

fsfod:
@ltratt how about that?

ltratt:
@fsfod Not all on one line please. Commits need a short overview sentence (max 80 chars, except in rare circumstances) then a blank line, then some more text below. In this case, for example, you might want to make the reader aware of why the `start` variable was in the code in the first place.

fsfod:
i only removed the start value because its not being used atm

ltratt:
That's a perfectly fine reason for removing it. Can you please put that into the commit message so that future readers know that? Otherwise they'll be left scratching their heads as to why it's disappeared and we don't want that ;)

fsfod:
@ltratt are you happy to merge it now?

fsfod:
@ltratt is that the last thing?

ltratt:
Yep!

PR #35, opened 2018-02-02
Reorgnize simplerunner in to more testable pieces

fsfod:
Refactored simplerunner to be more like a module to help make testing it easier and to help with running benchmarks in a child process where the code flow is much different from in process.

All functions should now be local or moved to a new runner lib table to avoid polluting the global table that could of conflict with any benchmark that might of used the same named global. 
The initialization of various modules and systems is now all done in a single function instead of being spread around the whole file where they would get instantly run when the file is loaded

### DIFF
--- a/simplerunner.lua
+++ b/simplerunner.lua
@@ -1,45 +1,9 @@
-local jit, jit_util, jit_dump, jitstats
+local jit, jit_util, jit_dump, jitstats, json
 local match, gmatch = string.match, string.gmatch
+local benchinfo, scaling
 
 local function add_package_path(basepath)
     package.path = string.format("%s/?/init.lua;%s/?.lua;%s/?/?.lua;%s", basepath, basepath, basepath, package.path)
-end
-add_package_path("lualibs")
-
-local json = require("json_nojit")
-
-local success
-success, jit = pcall(require, "jit")
-
-if not success or not pcall(jit.on) then
-    jit = nil
-else
-    success, jit_util = pcall(require, "jit.util")
-
-    if not success then
-        jit_util = nil
-    end
-end
-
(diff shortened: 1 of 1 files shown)


simplerunner.lua
```diff
@@ -57,35 +57,33 @@ function runner.init()
...
-            runner.jitfunc(startimer)
-            runner.jitfunc(stoptimer)
-        else
-            startimer, stoptimer = runner.maketimer(os.clock)
-            -- os.clock can't be jit'ed so try to keep timer a bit more stable
-            jit.off(startimer)
-            jit.off(stoptimer)
-        end
+    local timeloaded, timelib = pcall(require, "time")
+    -- jit.util is need so we can check the bytecode has changed for the timer functions being pre JIT'ed.
```
ltratt: "is needED"

ltratt:
I only have one minor comment as things stand. There are a couple of grammar issues in commit messages, but will they be squashed away eventually?

fsfod:
i would guess so if there not the first commit

ltratt:
OK, so in the first commit: "should of" -> "should have"; "might of" -> "might have"; sentence after "and get instantly run" needs rephrasing. You can squash (make sure you fix my other comment too please!) at the same time.

ltratt:
@fsfod What is the status of this?

fsfod:
@ltratt I did try to improve the messages a bit

fsfod:
But i did find 2 small bugs that you can see commits for

ltratt:
OK, the new commit and the messages look reasonable. Can you squash and wrap the commit messages to 80 chars at the same time please?

fsfod:

PR #36, opened 2018-02-05
Reset the hot counters at benchmark start

fsfod:
All the initialization code that simplerunner runs will of caused the shared hot counters to not be at there starting values before we run the benchmark. Try to keep things consistent by resetting them to there starting values right before we run the benchmark but after we've loaded it.

### DIFF
--- a/simplerunner.lua
+++ b/simplerunner.lua
@@ -231,6 +231,13 @@ function runner.runbench(name, count, scaling)
     io.write("Running "..name..": ")
     io.flush()
 
+    -- Reset all the hot counters to their starting values right before we run the benchmark
+    if jit then
+        -- lj_dispatch_update triggers a hotcount reset when the JIT is turned off and back on again.
+        jit.off()
+        jit.on()
+    end
+
     local run_iter = _G.run_iter
     for i = 1, count do
         io.write(".")


ltratt:
Can we read this value *before* the first benchmark, and then reset it to that in the future? What I'm worried about is that we're hardcoding in a value that might be quite different between different forks/versions of LuaJIT.

fsfod:
The only api to it is write only but i might have another way to do it by turning the JIT off and on again will reset the hot counts.

fsfod:
@ltratt how about that way then?

ltratt:
Looks good to me. I'm not sure we want the "alternative way" doc though -- since that number might not be accurate for some LuaJIT variants?

fsfod:
It looks like its been 56 since https://github.com/LuaJIT/LuaJIT/commit/8ae2f9feaaed874e01a87df6646242b80acceabb but i could change it to 'start("hotloop=DEFAULTIN-LJ_JIT.H")'

fsfod:
@ltratt are you happy with the code now?

ltratt:
Yep. If you squash, I can merge.

fsfod:

ltratt:
Sorry: need to "there" -> "their" in your commit message.

fsfod:
@ltratt fixed

fsfod:
@ltratt it was fixed again

ltratt:
Thanks!

PR #37, opened 2018-02-06
Add support for benchmark build scripts

fsfod:
The main build script was modified to run any build.sh scripts it finds in the benchmark directories and the **binary** Lua module search path now also updated for the benchmark directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,11 @@ write the following:
 }
 ```
 
+If your benchmark requires additional building / installation, you can add an 
+executable build.sh file in your benchmark directory. If present, this will be
+run during the initial build of the entire repository. Your build.sh file is 
+responsible for determining if previous builds are present and need to be 
+replaced etc.
 
 ## Benchmarking using Krun
 
--- a/build.sh
+++ b/build.sh
@@ -99,3 +99,10 @@ if [ ${BUILDALL} ]; then
   cp lua-vermelha/luav ${ljbins}/lua-vermelha/luav
 fi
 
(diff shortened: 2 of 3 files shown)


ltratt:
We have a failing test?

ltratt:
We also need to add a bit of text to the readme telling people that a) they can optionally add a `build.sh` file b) the file needs to determine itself if it actually needs to (re)compile anything.

fsfod:
how about 

> If your benchmark needs to build any extra binaries you can add a build.sh script
> to the benchmark's directory which gets automatically run from the main build.sh 
> script in the root directory

ltratt:
Good start. Maybe tweak it to "If your benchmark requires additional building / installation, you can add an executable `build.sh` file in your benchmark directory. If present, this will be run during the initial build of the entire repository. Your `build.sh` file is responsible for determining if previous builds are present and need to be replaced etc."?

fsfod:
we also never mentioned anything about running main build.sh

ltratt:
Good point. I'll make a PR to address that.

One other question: have you tested the running of a `build.sh` in a benchmark?

ltratt:
[I'll do that PR after this one is merged BTW.]

fsfod:
They don't get run in a benchmark. i have tested it runs build.sh scripts in benchmark directories

ltratt:
If you squash, I can merge.

fsfod:

fsfod:
@ltratt its not merged

ltratt:
Give me a chance ;)

PR #38, opened 2018-02-07
Run each benchmark in a new sub process

fsfod:
Run each benchmark in a new sub process that writes its results out to a temporary json file that the main process reads back in and prints the stats for it. To simplify debugging benchmarks there is an new option to run them in process like before.
There is also now a command line option to save all the benchmark results to a json file including the options were used to generate them which also includes benchmark scaling values.

### DIFF
--- a/.travis/platform.sh
+++ b/.travis/platform.sh
@@ -1,0 +1,15 @@
+if [ -z "${PLATFORM:-}" ]; then
+  PLATFORM=$TRAVIS_OS_NAME;
+fi
+
+if [ "$PLATFORM" == "osx" ]; then
+  PLATFORM="macosx";
+fi
+
+if [ -z "$PLATFORM" ]; then
+  if [ "$(uname)" == "Linux" ]; then
+    PLATFORM="linux";
+  else
+    PLATFORM="macosx";
+  fi;
+fi
--- a/simplerunner.lua
+++ b/simplerunner.lua
@@ -82,6 +82,7 @@ function runner.init()
     loaded = true
     oskind = runner.detectos()
     add_package_path("lualibs")
+    add_package_path("rocks/modules")
     runner.loadbenchinfo()
     
     success, jit = pcall(require, "jit")   
--- a/simplerunner.lua
+++ b/simplerunner.lua
@@ -1,6 +1,6 @@
 local jit, jit_util, jit_dump, jitstats, json
 local match, gmatch = string.match, string.gmatch
-local benchinfo, scaling, oskind
+local benchinfo, scaling, oskind, subprocess
 
 local function add_package_path(basepath)
     local ext
@@ -13,6 +13,8 @@ local function add_package_path(basepath)
     package.path = string.format("%s/?/init.lua;%s/?.lua;%s/?/?.lua;%s", basepath, basepath, basepath, package.path)
     package.cpath = string.format("%s/?.%s;%s/?/?.%s;%s", basepath, ext, basepath, ext, package.cpath)
 end
+
+ffi = require("ffi")
 
 local function table_filter(t, f)
     local result = {}
@@ -30,9 +32,26 @@ local function readfile(path)
         return false, msg
     end
 
-    local success, text = pcall(file.read, file, "*all")
-    file.close(file)  
-    return success, text
+    local result, err = file:read("*all")
+    file:close()
+    if not result then
+        return false, err
+    end
+    return true, result
+end
+
+local function writefile(path, data)
+    local file, msg = io.open(path, "w")
+    if not file then
+        return false, msg
+    end
+
+    local success, msg = file:write(data)
+    file:close()
+    if not success then
+        return false, msg
+    end
+    return true
 end
 
 local function loadjson(path)
@@ -315,21 +334,67 @@ function runner.run_benchmark_list(benchmarks, count, options)
 
     local package_path = package.path
     local package_cpath = package.cpath
-    
-    for i, name in ipairs(benchmarks) do
-        local times = runner.runbench(name, count, options.scaling or scaling[name])
-        local stats = runner.calculate_stats(times)
-        print(string.format("  Mean: %f +/- %f, min %f, max %f", stats.mean, stats.cinterval, stats.min, stats.max))
-        if jitstats then
-            jitstats.print()
-        end
-        -- Try to clean away the current benchmark and its data, so the behaviour of the GC is more predictable for the next benchmark.
-        collectgarbage("collect")
+    local results = {}
 
-        -- Restore the Lua module search path since we set it to the directory of the benchmark when we load it.
-        package.path = package_path
-        package.cpath = package_cpath
+    for i, name in ipairs(benchmarks) do
+        local scaling = options.scaling or scaling[name] or 1
+        local stats
+
+        if options.inprocess then
+            local times, jstats = runner.runbench(name, count, scaling)
+            stats = runner.calculate_stats(times)
+            stats.times = times
+            print("  " .. runner.fmtstats(stats))
+            if jitstats then
+                stats.jitstats = jstats
+                jitstats.print()
+            end
+        else
+           runner.runbench_outprocess(name, count, scaling, options)
+        end
+
+        if options.inprocess then
+            -- Try to clean away the current benchmark and its data, so the behaviour of the GC is more predictable for the next benchmark.
+            collectgarbage("collect")
+
+            -- Restore the Lua module search path since we set it to the directory of the benchmark when we load it.
+            package.path = package_path
+            package.cpath = package_cpath
+        end
     end
+end
+
+function runner.runbench_outprocess(name, count, scaling, options)
+    local lc = require("luachild")
+    local read, write = lc.pipe()
+
+    local p = lc.spawn{
+        command = arg[-1],
+        args = {
+            arg[0], "--childprocess", name, count, scaling, unpack(arg, 1) 
+        },
+    }
+ 
+    local cmdret = p:wait()
+    if cmdret ~= 0 then
+        error("Out of process benchmark execution failed status = "..cmdret)
+    end
+end
+
+function runner.subprocess_run(benchmark, count, scaling, parent_options)
+    scaling = tonumber(scaling)
+    count = tonumber(count)
+    subprocess = true
+    runner.processoptions(parent_options)
+
+    local times, jstats = runner.runbench(benchmark, count, scaling)
+    local stats = runner.calculate_stats(times)
+    print("  " .. runner.fmtstats(stats))
+    if jitstats then
+        jitstats.print()
+    end
+
+    os.exit(0)
 end
 
 function runner.calculate_stats(times)
@@ -358,6 +423,10 @@ function runner.calculate_stats(times)
     -- 95% confidence interval
     stats.cinterval = 1.960 * stats.stddev/math.sqrt(count)
     return stats
+end
+
+function runner.fmtstats(stats)
+    return string.format("Mean: %f +/- %f, min %f, max %f", stats.mean, stats.cinterval, stats.min, stats.max)
 end
 
 -- Command line parsing system mostly copied from dynasm.lua in LuaJIT
@@ -404,17 +473,18 @@ function opt_map.help()
 
 Usage: simplerunner [OPTION] [<benchmark>] [<count>]
 
-  -h, --help           Display this help text.
-  -b, --bench name     Name of benchmark to run.
-  -c, --count num      Number of times to run the benchmarks.
-  -s, --scaling num    Number of interations to run the benchmarks.
-  -e, --exclude name   Exclude a benchmark from being run.
+  -h, --help            Display this help text.
+  -b, --bench name      Name of benchmark to run.
+  -c, --count num       Number of times to run the benchmarks.
+  -s, --scaling num     Number of interations to run the benchmarks.
+  -e, --exclude name    Exclude a benchmark from being run.
 
   --jitstats           Collect and print jit statisitcs from running the benchmarks.
   --jdump options      Run LuaJIT's jit.dump module with the specifed options.
   --benchloadjitstats  Also collect jitstats when loading a benchmark
   --nogc               Run the benchmarks with the GC disabled.
   --norandomize        Don't randomize the run order of the benchmarks.
+  --inprocess          Run benchmarks in the main process instead of executing them in a child process.
 ]]
     os.exit(0)
 end
@@ -428,6 +498,7 @@ function opt_map.jitstats() g_opt.jitstats = true end
 function opt_map.benchloadjitstats() loading_jitstats = true end
 function opt_map.nogc() g_opt.nogc = true end
 function opt_map.norandomize() g_opt.norandomize = true end
+function opt_map.inprocess() g_opt.inprocess = true end
 function opt_map.jdump(args)
     local options = optparam(args)
     local outfile = options:find(",")
@@ -455,7 +526,6 @@ end
 -- Parse arguments.
 function runner.parse_commandline(args)
     -- Default options.
-    g_opt.count = 30
     g_opt.benchmarks = {}
     g_opt.excludes = {}
     -- Process all option arguments.
@@ -483,6 +553,8 @@ function runner.parse_commandline(args)
         end
     end
 
+    g_opt.count = g_opt.count or 30
+
     local benchmarks
     if #g_opt.benchmarks > 0 then
         benchmarks = g_opt.benchmarks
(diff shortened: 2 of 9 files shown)


simplerunner.lua
```diff
@@ -35,6 +35,21 @@ local function readfile(path)
     return success, text
 end
 
+local function writefile(path, data)
```
ltratt: This function appears to sometimes return `(bool, String)` and sometimes just `bool`. That seems like a recipe for confusion...

simplerunner.lua
```diff
@@ -317,18 +332,65 @@ function runner.run_benchmark_list(benchmarks, count, options)
...
-        -- Restore the Lua module search path since we set it to the directory of the benchmark when we load it.
-        package.path = package_path
-        package.cpath = package_cpath
+        print("  " .. runner.fmtstats(stats))
+
+        if options.inprocess then
+            -- Try to clean away the current benchmark and its data, so the behaviour of the GC is more predictable for the next benchmark.
+            collectgarbage("collect")
+
+			-- Restore the Lua module search path since we set it to the directory of the benchmark when we load it.
```
ltratt: Random indentation?

simplerunner.lua
```diff
@@ -317,18 +332,65 @@ function runner.run_benchmark_list(benchmarks, count, options)
...
+    local times, jstats = runner.runbench(benchmark, count, scaling)
+    local stats = runner.calculate_stats(times)
+    stats.times = times
+    stats.jitstats = jstats
+
+    local data = json.encode(stats)
+    local success, msg = writefile(resultsfile, data)
+
+    if not success then
+      print("Failed to save results", msg)
```
ltratt: 2-space indentation.

simplerunner.lua
```diff
@@ -415,6 +481,7 @@ Usage: simplerunner [OPTION] [<benchmark>] [<count>]
   --benchloadjitstats  Also collect jitstats when loading a benchmark
   --nogc               Run the benchmarks with the GC disabled.
   --norandomize        Don't randomize the run order of the benchmarks.
+  --inprocess          Run bench in the main process instead of executing them in a child process
```
ltratt: "Run bench*marks*"

simplerunner.lua
```diff
@@ -375,11 +376,18 @@ function runner.runbench_outprocess(name, count, scaling, options)
...
-function runner.subprocess_run(benchmark, count, scaling, resultsfile)
-    print('RUNBENCH', benchmark, count, scaling, resultsfile);
+function runner.subprocess_run(benchmark, count, scaling, optionfile, resultsfile)
+    print('RUNBENCH', benchmark, count, scaling, optionfile, resultsfile);
     scaling = tonumber(scaling)
     count = tonumber(count)
 
+    local success, parent_options = loadjson(optionfile)
+    if not success then
+      print("Failed to read parents options:", parent_options)
```
ltratt: 2-space indentation.

simplerunner.lua
```diff
@@ -551,6 +563,32 @@ function runner.parse_commandline(args)
...
+        -- If the optional file doesn't specify a list of benchmarks allow the command line options set them
+        if #file_benchmarks > 0 then
+            g_opt.benchmarks = file_benchmarks
+        end
+
+        if #file_excludes > 0 then
+            g_opt.excludes = file_excludes
+        end
+
+        -- Merge options from the file with command line. Conflicting command line options will get overwritten for simplicity.
```
ltratt: I don't understand the first sentence.

simplerunner.lua
```diff
@@ -630,6 +668,9 @@ else
         benchmarks = benchlist
     end
     benchmarks = runner.filter_benchmarks(benchmarks)
+    -- Only save the benchmarks were going to run in our last used options
```
ltratt: "were" -> "we're"

simplerunner.lua
```diff
@@ -627,6 +628,12 @@ function runner.parse_commandline(args)
 end
 
 function runner.processoptions(options)
+
```
ltratt: Extraneous blank line.

simplerunner.lua
```diff
@@ -627,6 +628,12 @@ function runner.parse_commandline(args)
...
+    -- Don't pointlessly run these options in the main process if were running benchmarks in a child process
```
ltratt: "were" -> "we're"

simplerunner.lua
fsfod: The non error throwing way of reporting errors in Lua is to return false and the error message or true, and the expected result but theres not always something to return

ltratt: Surely it's better to always return a pair, with null (or similar) in the second position if there's nothing better?

simplerunner.lua
fsfod: @ltratt how about "Merge values from the option file with command line options. Conflicting command line options will get overwritten for simplicity."

ltratt: Ok, now I get it. But I think conflicting options should probably lead to an error, otherwise people are going to do all sorts of weird things. Thoughts?

fsfod: @ltratt I wanted to have like an overlay system where the base options could be set by an option file but the user could override some value on the command line. 
I started out with the commandline overriding the options file but i couldn't tell if a command line option was set because there are default values in the optional table already but i think i can remove them and go back to command line always overriding

ltratt: I must admit that I think this is too much complexity, at least until we have a concrete use case. Can we simplify this stuff please?

vext01:
Silly question on my part, but why do you need this?

fsfod:
It will help randomize the environment the benchmarks run in and eliminate any chance of side effects from running benchmarks one after another in the same environment which is probably already happening because of the same modules being used between benchmarks

simplerunner.lua
fsfod: The use case was the same as krun overlay files to quickly experiment from a base setup and to customize to an particular environment but did simplify from what i started

ltratt:
OK, it's now clear that there are two completely different parts to this PR. One runs things in a subprocess (I like this). The other has cache files and other options. The latter I'm a lot less sure about -- it certainly needs better motivating and probably a better design.

So what I'd like to do is to chop this PR down a lot and make it *only* about running things in a subprocess. We can then discuss the design of the other part separately and, if we think it's still a good idea, open a new PR.

I know this is a fair bit of work, but I think it's dangerous to proceed with this mixed-PR as is. Sound like a plan,?

fsfod:
The options files are used as part of running the subprocess. The command line option to load them just seem liked a natural extension of it. The saving results to a file system is used to get the results back from the sub process

ltratt:
OK, as a first step, can you outline the overall design more clearly. Then we can work out what needs to be split out. [The command-line option stuff certainly shouldn't be part of this PR from what I can see so far, but maybe I'm wrong.]

fsfod:
The sub process is executed with a fixed set of arguments [here](https://github.com/softdevteam/lua_benchmarking/pull/38/files#diff-77269539bc7b7a261edd81801f33fc3bR382 ) which is the benchmark name, iter count and scaling. The other values passed are options file and the path to write the results of the benchmark. The non benchmark specific options like nogc, collect jit stats, load LuaJIT's jdump are loaded from the options file

ltratt:
A few initial questions:

  * Why do we need an options file?
  * Could we parse the output of the child process' stdout/stderr rather than writing to a file?

fsfod:
I wanted to avoid all the issues with passing all the other options on the command line like re-escaping them and making the argument parser run in two modes. If i wanted to pass them as fixed position i would have to update the code in many places everytime i want to add a new option.

I thought about this but we are printing the progress dot and Lua's popen api is limited to only stderr or stdout and the jitstats results is not just a list of numbers.

fsfod:
> If you use execve (or equivalent), there's no problem with escaping arguments (escaping is purely a problem when you're at the human-level shell). 

I would have to use FFI to call that and I thought you wanted this still work on normal Lua. What i had works without any extra modules or ffi calls. I was trying to aim to run less Lua code in the child process before the benchmark but it probably ends up the same with json or arg parsing.

> Can you explain why you'd need two arg parser modes?

I was thinking in terms of writing out the parsed options as the command line because some wouldn't make sense in the child process but they would be normalized to different values/names in the table, but i guess copying the array of arg strings could work.

ltratt:
A module like https://github.com/mah0x211/lua-process looks like it provides the necessary functionality.

fsfod:
@ltratt it is simplified down now and I was looking at https://github.com/pocomane/luachild before but its not needed now.

ltratt:
Can you quickly sketch out the design of the simplified version?

fsfod:
The child process just writes the stats out and extra options are still passed as json file for simplicity

ltratt:
Let's use `execve` (or similar; there are all sorts of nice wrappres) and pass the arguments directly: using json is just extra code we have to maintain.

fsfod:
json is already being used for benchmark info

fsfod:
Using execve on windows i get screwed up output with the prompt being written in the middle of the progress output and i have to hit enter to get my prompt back after it finish. I can't even cancel it while its in process with ctl-c

ltratt:
Does that happen on bencher8?

fsfod:
I mean its a native api on Linux. On my Linux machine it doesn't have those issues

ltratt:
Is that a Windows bug then? Or a buggy Lua library?

fsfod:
I was calling the api directly through LuaJIT's ffi and os.execute calls system()

ltratt:
Can you use one of the luarocks library that bundles this stuff up? It's difficult to get the dance with child processes and stdin/stdout and so on right. Let's not fear using external libraries.

simplerunner.lua
```diff
@@ -315,20 +332,62 @@ function runner.run_benchmark_list(benchmarks, count, options)
...
+            -- Restore the Lua module search path since we set it to the directory of the benchmark when we load it.
+            package.path = package_path
+            package.cpath = package_cpath
+        end
+    end
+end
+
+function runner.runbench_outprocess(name, count, scaling, options)
+    local optionsfile = "lastrun.json"
+    local cmdret = os.execute(string.format('%s %s @RUNBENCH@ "%s" %d "%s" "%s"', arg[-1], arg[0], name, count, scaling, optionsfile))
```
ltratt: This isn't a safe way of calling a child process. We need to use `execve` or equivalent.

fsfod:
I'm testing out using luachild but the way it builds in luarocks doesn't really work well

ltratt:
OK, let's stick at it. If that doesn't work, there's almost bound to be another library which does what we want.

fsfod:
@ltratt  This is a bit rough but i turned luachild in to a ffi module and used that for process creation along with testing out std pipe redirection which works on both Linux and windows.

simplerunner.lua
```diff
@@ -363,10 +366,35 @@ end
...
+    local p = lc.spawn{ 
+        args = {
+            arg[-1], arg[0], "@RUNBENCH@", name, count, scaling, unpack(arg, 1) 
+        },
+        stdout = write,
+    }
+
+    local s = read:read("l")
+    print(s)
+    -- Close out write side read doesn't when the child process exits
```
ltratt: I can't parse this sentence.

simplerunner.lua
```diff
@@ -363,10 +366,35 @@ end
...
+    write:close()
+    
+    local c = read:fgetc()
+    while c do
+        io.stdout:write(c)
+        io.stdout:flush()
+        c = read:fgetc()
+    end
 
+    --local json = stderr_out:read("l")
```
ltratt: Dead code?

ltratt:
It doesn't make sense to include the luachild source code -- let's please download it in `build.sh` (or tell someone how to install it with luarocks).

fsfod:
so only raptorjit failing because of its lua shared library name

ltratt:
Can you expand a bit on what the RaptorJIT problem is please?

fsfod:
well the error is  `rocks/modules/luachild.so: undefined symbol: lua_gettop` either the lua shared library is not loaded or the symbol wasn't exported from it. lua_gettop is part of the core api as well so shouldn't need any special compile flags

fsfod:
it looks like it was exported correctly
```
readelf -Ws ./builds/raptorjit/libluajit.so | grep gettop
192: 0000000000018bd0    13 FUNC    GLOBAL DEFAULT   12 lua_gettop
```

fsfod:
@ltratt well it looks like it raptorjit was statically linked to its lua library

fsfod:
@ltratt sqaushed

ltratt:
I don't think `luachild_ffi.lua` should be a part of this (that's why we're using rocks, right?). Also, I thought your plan was to use `--inprocess` on Travis, meaning that we can drop the `.travis` directory stuff?

fsfod:
@ltratt  .travis directory is for getting a newer version of LuaRocks for travis

ltratt:
I thought what you were saying yesterday is that you could avoid that if you used `--inprocess`? Or maybe I misunderstood?

fsfod:
It was to work around RaptorJIT statically linking the lualib that --inprocess fixed

ltratt:
Aha! OK, got it.

Let's still please get rid of the luachild file though.

fsfod:
@ltratt i fixed a few things i just noticed and removed luachild_ffi

fsfod:
@ltratt ping

ltratt:
What's the point of the Windows `.bat` file? I thought Windows could just use the normal Unix shell scripts?

fsfod:
@ltratt Nope WSL(Windows Subsystem for Linux) basically emulates Linux at the system call layer only the newest version of windows has it which is only about 40% of the windows ecosystem. There is Cygwin and MinGW but there is likely to be lots of build issues and missing libs with trying to use them.
The bat just compiles native windows binaries of our LuaJIT and the needed LuaRocks using the native microsoft compiler, which will just work on all the main windows versions people use.

ltratt:
OK, but how had you built things on Windows up until this point?

fsfod:
Well it was adhoc mess before i would build stuff with the windows native compiler and separately copy stuff around try to recreate file layout that the Linux build script would create. This didn't matter as much before when we weren't using LuaRocks

ltratt:
OK, let's break out the Windows stuff into a separate PR so it can be considered sensibly.

fsfod:
@ltratt Can i squash down the fixes now?

.travis/platform.sh
```diff
@@ -0,0 +1,15 @@
+if [ -z "${PLATFORM:-}" ]; then
+  PLATFORM=$TRAVIS_OS_NAME;
```
ltratt: Inconsistent indentation.

simplerunner.lua
```diff
@@ -315,23 +335,69 @@ function runner.run_benchmark_list(benchmarks, count, options)
...
 end
 
+function runner.runbench_outprocess(name, count, scaling, options)
+    local lc = require("luachild")
+    local read, write = lc.pipe()
+
+    local p = lc.spawn{
+        command = arg[-1],
+        args = {
+            arg[0], "@RUNBENCH@", name, count, scaling, unpack(arg, 1)
```
ltratt: Is `@RUNBENCH@` a magic marker? Wouldn't it be cleaner to have a "hidden" option "--child" or something like that?

.travis/platform.sh
fsfod: I imported that file unmodified from that Lua travis repo

ltratt: Point taken.

simplerunner.lua
fsfod: It made things simpler to have those arguments in fixed position they don't make any sense in the parent process but i could just change @RUNBENCH@ to --child or --runbench but it would still just be a marker. I also branch on it to initialize the process differently in the child process

ltratt: I'm fine with it being in a fixed position, but I think having it as a `--` option (whatever we call it) will be cleaner. [I'm also nervous about how some weird shell will interpret `@` in a file name!]

fsfod: i changed it to --childprocess

.travis/platform.sh
fsfod: I did modify setup_lua.sh though to cut it down to just building and installing Lua rocks

ltratt: If you didn't modify this file at all, I think it's OK to leave the indentation as-is.

ltratt:
OK... time to squash :)

fsfod:

ltratt:
Let's get a bit more detail into the first commit message please. [e.g. motivation for why we want this feature.]

In the second commit, I can't parse the sentence "Override the LuaRocks config" -- add some commas or fullstops? [And I think it should be "headers/libs" with a forward, not a back, slash.]

ltratt:
"It also helps randomized the environment  the benchmarks are run in more." "randomized" -> "random". Double space before "more".

"install to a well know" -> "well known"

fsfod:
@ltratt are you happy to merge this now?

PR #39, opened 2018-02-07
Add jitlog instrumentation to our krun runner

fsfod:
Update our LuaJIT iterations_runner.lua to collect jitlogs for benchmarks when instrumentation is enabled and fix luajit.krun for the api changes in krun for the new instrumentation system

### DIFF
--- a/iterations_runner.lua
+++ b/iterations_runner.lua
@@ -84,19 +84,31 @@ local BM_benchmark = arg[1]
 local BM_iters = tonumber(arg[2])
 local BM_param = tonumber(arg[3])
 local BM_debug = tonumber(arg[4]) > 0
-local BM_instrument = tonumber(arg[5]) > 0
+local BM_instrument = #arg >= 5
+local BM_instdatadir = arg[5]
+local BM_key = arg[6]
+local BM_pexecidx = arg[7]
 
-if BM_instrument then
-    jitstats = require("jitstats")
-end
-
-if #arg ~= 5 then
-    io.stderr:write("usage: iterations_runner.lua <benchmark> <# of iterations> " ..
-                    "<benchmark param> <debug flag> <instrument flag>")
+if #arg ~= 4 and #arg ~= 7 then
+    io.stderr:write("usage: iterations_runner.lua <benchmark> " ..
+                    "<# of iterations> <benchmark param>\n           " ..
+                    "<debug flag> [instrumentation dir] [key] " ..
+                    "[key pexec index]\n\n")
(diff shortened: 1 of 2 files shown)


iterations_runner.lua
```diff
@@ -84,20 +84,28 @@ local BM_benchmark = arg[1]
...
+local BM_instrument = arg[5]
+local BM_key = arg[6]
+local BM_pexecidx = arg[7]
 
-if BM_instrument then
-    jitstats = require("jitstats")
-end
-
-if #arg ~= 5 then
+if #arg ~= 7 and #arg ~= 4 then
```
vext01: The usage string is incorrect.

iterations_runner.lua
```diff
@@ -147,7 +151,7 @@ for BM_i = 1, BM_iters, 1 do
         io.stderr:write(string.format("[iterations_runner.lua] iteration %d/%d\n", BM_i, BM_iters))
     end
     if BM_instrument then
-        jitstats.getsnapshot(stats_start)
+        jitlog.addmarker("BEGIN")
```
vext01: Is it worth including an iteration number in your markers? Just for sanity.

vext01:
Is it worth bringing the iteration runner as close as possible to the one in Krun?

fsfod:
Well it was the based on the in krun i think the only difference was for the instrumented mode

vext01:
Sure, but the handling of the number of args, for example, differs.

It's up to you...

iterations_runner.lua
fsfod: It was half place holder but adding a number to the string makes it a bit more of a hassle to parse in the reader. The flag bits are unused for the marker so i could stuff the count in there.

fsfod:
@vext01 I made the code a bit closer to kruns version. i don't know why you separated the two check argument checks in yours. You can safely access past the end of a table Lua you just always get back nil not something weird like i assume undefined in javascript.

ltratt:
Let's make it as close as is practical, otherwise we're going to trip ourselves up...

vext01:
Thanks.

The reason the arg handling is that way, is that I was implementing this for all of the runners and wanted the same logic across all.

I think we are good to squash, unless Laurie has any comments?

fsfod:

ltratt:
Unless I'm missing out on something, I didn't say this was ready for squashing. Can you show me the commit(s) I missed?

fsfod:
well it was @vext01 said to squash i just thought i you would get back quicker. There were no new commits

PR #40, opened 2018-02-15
Add a build script for windows

fsfod:
The batch script builds both a normal and a GC64 build of LuaJIT from our repo using LuaJIT's windows build bat script which is much more limited than the Linux makefile so no dualnum builds. It then builds the luachild module with LuaRocks.

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -1,0 +1,65 @@
+@echo off
+
+setlocal enableextensions enabledelayedexpansion
+
+if not defined INCLUDE (
+  echo No MSVC env vars set attempting to find vcvars.bat using system installed vswhere.exe
+  set vswhere="%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe"
+  
+  if exist !vswhere! (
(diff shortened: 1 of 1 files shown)


ltratt:
Dumb question: would it be neater to use a Unix shell script to drive the Windows build? [Windows batch files are not much fun to look at.]

fsfod:
Well its not something most people will have installed or actually in there PATH

ltratt:
I thought it was a default in Win10?

fsfod:
Its a feature you have to install and once you invoke it your in a different filesystem

ltratt:
OK.

Let's call it `build.bat` for consistency with `build.sh` and then I think we can merge it.

fsfod:
I did originally have it as that but changed it so both platform build scripts names would fully autocomplete and less chance to mix them up

ltratt:
Let's rename it please.

fsfod:
@ltratt can i squash now?

ltratt:
Yes, please squash.

fsfod:
@ltratt squashed

PR #41, opened 2018-03-05
Add the benchmark directory to the Lua module path

fsfod:
Update the Krun runner to add the directory that the benchmark script is contained as a Lua module search path. This fixes luafun when run under Krun.

### DIFF
--- a/iterations_runner.lua
+++ b/iterations_runner.lua
@@ -40,6 +40,28 @@ local ffi = require("ffi")
 local json = require("json_nojit")
 local jitstats
 
+-- @param addtostart Pass true to insert the search path at the start of the module search path list.
+function add_luapackage_path(basepath, addtostart)
+    assert(type(basepath) == "string" and #basepath > 0)
+
+    if addtostart then
+        package.path = string.format("%s/?.lua;%s/?/init.lua;%s", basepath, basepath, package.path)
+    else
+        package.path = string.format("%s;%s/?.lua;%s/?/init.lua", package.path, basepath, basepath)
+    end
+end
+
+-- @param addtostart Pass true to insert the search path at the start of the module search path list.
+function add_cpackage_path(basepath, addtostart)
+    assert(type(basepath) == "string" and #basepath > 0)
+
+    if addtostart then
+        package.cpath = string.format("%s/?.so;%s/?/?.so;%s", basepath, basepath, package.cpath)
+    else
+        package.cpath = string.format("%s;%s/?.so;%s/?/?.so", package.cpath, basepath, basepath)
+    end
+end
+
 function emit_per_core_measurements(name, num_cores, tbl, tbl_len)
     io.stdout:write(string.format('"%s": [', name))
 
@@ -96,6 +118,13 @@ if #arg ~= 4 and #arg ~= 7 then
                     "[key pexec index]\n\n")
     io.stderr:write("Arguments in [] are for instrumentation mode only.\n")
     os.exit(1)
+end
+
+-- Add the benchmarks directory as a Lua module search path
+local benchdir, scriptname = string.match(BM_benchmark, "^(.-)[/]?([^/]*)$")
+if benchdir and scriptname then
+    add_luapackage_path(benchdir, true)
+    add_cpackage_path(benchdir, true)
 end
 
 if BM_instrument then


ltratt:
What's the "highpriority" thing fixing?

fsfod:
It insert the directory at start of the module search path string so Lua VM searches for modules in that directory first.

ltratt:
Yes, I realised that bit :) But why is this needed sometimes and not others?

fsfod:
Some benchmarks are going to use it as well and the ordering matters for them

ltratt:
Are you talking about existing benchmarks or new benchmarks?

iterations_runner.lua
```diff
@@ -40,6 +40,34 @@ local ffi = require("ffi")
 local json = require("json_nojit")
 local jitstats
 
+function add_luapackage_path(basepath, highpriority)
+    assert(type(basepath) == "string" and #basepath > 0)
+
+    if string.sub(basepath, -1) == "/" then
```
ltratt: This isn't fully safe: for example `///blah` is a valid basepath. Is there a Lua library that manipulates paths that we can use? Getting this stuff right, and portable, is harder than it seems...

iterations_runner.lua
```diff
@@ -40,6 +40,34 @@ local ffi = require("ffi")
 local json = require("json_nojit")
 local jitstats
 
+function add_luapackage_path(basepath, highpriority)
```
ltratt: Rather than `highpriority` can we call it something like `addtoend` or `addtostart` or something? And let's document that it's a boolean above the function.

iterations_runner.lua
fsfod: this is why i asked for krun to provide it

ltratt: How do we know what comes out of Krun is fully normalised? Let's not manipulate paths as strings -- it's just a recipe for disaster.

fsfod: I mean for krun to pass the script directory path as a argument to the runner

ltratt: I understand that. But you can't guarantee it's normalised. So let's please use a library to fix this.

iterations_runner.lua
fsfod: i didn't know what to really call it that wasn't really long winded or provided more guarantee than what it does.

ltratt: Let's go with `addtostart` then. `highpriority` makes promises that it can't quite guarantee to keep. Let's add a docstring (or whatever the Lua equivalent is) to the function so that we document what's going on.

fsfod: @ltratt I had just added it as plain comment to the function header

ltratt: Is that the standard Lua way of doing things?

fsfod: @ltratt well there is a thirdparty system called LuaDoc that is more the structured kind that you might be thinking of

iterations_runner.lua
```diff
@@ -98,6 +120,14 @@ if #arg ~= 4 and #arg ~= 7 then
     os.exit(1)
 end
 
+-- Add the benchmarks directory as a Lua module search path
+local slashidx = string.find(BM_benchmark, "/[^/]*$")
```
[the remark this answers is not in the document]
ltratt: OK, so this is where we definitely need some sort of sanity-check assert, as this is definitely fragile (e.g. if Krun hands us `/a/b/c//` this will fail).

ltratt:
@fsfod Ping.

fsfod:
it does not

ltratt:
OK, then let's at least do an assert along the lines of "check that the substring I generated is a directory".

fsfod:
@ltratt is there anything else you want before i squash?

iterations_runner.lua
```diff
@@ -121,8 +121,9 @@ if #arg ~= 4 and #arg ~= 7 then
 end
 
 -- Add the benchmarks directory as a Lua module search path
-local benchdir = string.match(BM_benchmark, "^(.-)[/]?([^/]*)$")
-if benchdir then
+local benchdir, scriptname = string.match(BM_benchmark, "^(.-)[/]?([^/]*)$")
+if benchdir and scriptname then
+    assert(benchdir:sub(#benchdir, -1) ~= "/")
```
ltratt: Is there a Lua function that says "is this path $p$ a dir". e.g. in Python it would be `os.isdir(p)`. What we don't want to do is add something to the search path that isn't a dir: it'll look like it works, but silently fail in a confusing way.

fsfod: There is pretty much no built-in file system API only basic file IO

ltratt: There's no way using the standard library to do this?

fsfod: well you can check that a file exists by trying to open it in Lua but i don't know anyway to check a that directory exists with the built-in api.

ltratt: Can you have a look around to see if this is possible? If it absolutely isn't, then there isn't much we can do about it (even though I'm a bit stunned...).

fsfod: The only ways seem to be rename the dir to its self with os.rename(dirname, dirname) and check for an error or execute a shell command like 'cd' and check the result

ltratt: Renaming a directory? Erp. No, we definitely don't want to do that! OK, so Lua's standard library sucks in this regard :( Oh well. Let's remove the assert -- it's not useful as it stands. We'd better just cross our fingers that we don't encounter problems...

fsfod:
@ltratt can i squash now?

ltratt:
Yes, please do.

fsfod:

PR #42, opened 2018-03-09
Shutdown and bring up the network interface in the bencher.krun file

fsfod:
I just used the same systemd cmds in warmup.krun from warmup_experiment to shutdown the network when krun runs the benchmarks and bring them up after.

### DIFF
--- a/bencher.krun
+++ b/bencher.krun
@@ -2,7 +2,15 @@ execfile("luajit.krun", globals()) # inherit settings from the main experiment
 
 MAIL_TO = ["fsfod11@gmail.com"]
 
+PING_HOST = "bencher8.soft-dev.org"
+
 PRE_EXECUTION_CMDS = [
+    # Stop network first. If an interface accepts a DHCP lease during one
+    # of the later commands below, it can cause the command to be
+    # "cancelled" by systemd. Bringing the network itself down can fail in
+    # the same way, so keep trying (with sleeps between).
+    "while true; do sudo systemctl stop networking; sleep 5; ping -q -c 10 %s || break; done" % PING_HOST,
+    
     "sudo systemctl stop apt-daily.timer",
     "sudo systemctl --runtime disable apt-daily.timer",
 
@@ -19,6 +27,10 @@ PRE_EXECUTION_CMDS = [
 ]
 
 POST_EXECUTION_CMDS = [
+    # The network doesn't always come up properly on debian. We keep trying
+    # until we can ping a host on the network.
+    "while true; do ping -c 3 -q %s && break; sudo systemctl stop networking; sleep 5; sudo systemctl start networking; done" % PING_HOST,
+    
     "sudo systemctl start systemd-timesyncd || true",
     "sudo systemctl start cron || true",
 


ltratt:
I'm fine with this, but I think @vext01 is best placed to deal with this PR, so I'll leave him to it.

fsfod:
@vext01 ping

ltratt:
Did we reach a resolution on b9?

fsfod:
well it seems that loop was a terrible idea in the bring up

ltratt:
?

fsfod:
thats what bencher9 is stuck in 
`while true; do ping -c 3 -q %s && break; sudo systemctl stop networking; sleep 5; sudo systemctl start networking; done" % PING_HOST`

ltratt:
What host is it trying to ping? Can b9 ping that under normal circumstances?

fsfod:
it could ping it ping bencher 9 when i tried normally

ltratt:
@vext01 Any thoughts?

ltratt:
@fsfod Ah, you didn't answer my question. I asked what host is being pinged. Hopefully b9 isn't pinging itself!

bencher.krun
```diff
@@ -2,7 +2,15 @@ execfile("luajit.krun", globals()) # inherit settings from the main experiment
 
 MAIL_TO = ["fsfod11@gmail.com"]
 
+PING_HOST = "bencher8.soft-dev.org"
```
vext01: ^ It should be pinging bencher8.

fsfod:
well it also looks like systemctl start networking was doing nothing as well

vext01:
Ah, I saw that yesterday on bencher9 also. I think google is your friend at this point. Perhaps there is an update that fixes this?

ltratt:
Maybe Debian changed something to do with networking in a recent update?

vext01:
Certainly possible, although if past experience is worth anything, it's likely that it's just flaky.

Can you look into it please Tom?

fsfod:
did @vext01 do any hand editing of /etc/network/intefaces before i start changing it?

ltratt:
This has some possible pointers https://www.reddit.com/r/debian/comments/7nw98g/seriously_restarting_network_interfaces/

fsfod:
Adding `auto eno1` to /etc/network/interfaces fixed systemd not starting the network interfaces up again

vext01:
OK, so time for a re-test @fsfod ?

fsfod:
@vext01 well the network still didn't come back up

fsfod:
well maybe i guess it look much longer than i thought it would

ltratt:
Can you be clear about whether it works and what changes we might need to make to the Krun script please?

vext01:
So it works?

fsfod:
I wasn't sure if it was working before if it was stuck in a reboot loop but it did finish correctly and brought the network backup

ltratt:
I'm still unclear which variant you're talking about. Can you push a commit that fixes the problem? [If a commit is needed.]

fsfod:
i didn't change anything on the krun side. The fix to  /etc/network/interfaces was enough to get it to work with this pull request

vext01:
I think Tom is saying that it does work now. It shouldn't be a reboot loop, as Krun won't reboot until all of the POST_EXEC commands (one of which is bringing up the network) have finished.

ltratt:
So we don't have to change anything to this PR?

fsfod:
Well i'm not sure if we should pick a different ping host

fsfod:
Is bencher8 still going to be occasionally used for other krun experiments that reboot?

ltratt:
No, b8 is never used for rebooting experiments.

fsfod:
I guess theres nothing else to change then

ltratt:
I think we're ready to merge then. @fsfod @vext01?

fsfod:
ready to merge

PR #43, opened 2018-05-14
Convert the rest LuaJIT testsuite benchmarks to run under krun

fsfod:
Add the rest of the benchmarks from the LuaJIT testsuite repo https://github.com/LuaJIT/LuaJIT-test-cleanup/tree/master/bench.
The newly added benchmarks are:
array3d, coroutine_ring, euler14 nsieve, nsieve_bit, life, mandelbrot, mandelbrot_bit, ray, meteor, pidigits, recursive_ack, recursive_fib, partialsums, scimark_fft, scimark_lu, scimark_sor, scimark_sparse, scimark_mc, revcomp, knucleotide.


iterations_runner.lua
```diff
@@ -40,6 +40,18 @@ local ffi = require("ffi")
 local json = require("json_nojit")
 local jitstats
 
+local devnull
+if package.config:sub(1,1) == "\\" then
+  -- Use nul on windows https://stackoverflow.com/questions/313111/is-there-a-dev-null-on-windows
```
ltratt: Does that mean that one can never have a file called `nul` on Windows? That seems weird!

ltratt:
@ebarrett would you be able to take a look at this? I'm on a low bandwidth link, so it's difficult for me to look at all of this...

iterations_runner.lua
fsfod: Its considered a reserve name so its not something you can create though any normal means.

benchmarks/array3d/bench.lua
```diff
@@ -0,0 +1,61 @@
...
+    image = {}, -- Preferably use a fixed-type, pre-sized array here.
+    set = array_set,
+    points = array_points,
+  }
+end
+
+local packed = true   -- Packed image or flat
+
+function run_iter(scaling)
+  local dim = scaling -- Array dimension dim^3
```
vext01: Would it make more sense to have linear scaling? Or is that not possible?

benchmarks/chamenos/bench.lua
```diff
@@ -0,0 +1,69 @@
...
+        threads[i] = nil
+      end
+    end
+  until false
+end
+
+local threads 
+
+function run_iter(scaling)
+    N = scaling
```
vext01: If I were doing this, I'd probably have tried to kill the global state, but it's up to you.

vext01: There are other examples in other benchmarks too.

benchmarks/life/bench.lua
```diff
@@ -0,0 +1,115 @@
...
+    thisgen:evolve(nextgen)
+    thisgen,nextgen = nextgen,thisgen
+    write("\027[H")	-- ANSI home cursor
+    thisgen:draw()
+    write("Life - generation ",gen,"\n")
+    gen=gen+1
+    if gen > n then
+      break 
+    end
+    --delay()		-- no delay
```
vext01: Should we just kill this line?

benchmarks/mandelbrot/bench.lua
```diff
@@ -0,0 +1,28 @@
+local write, char, unpack = io.write_devnull, string.char, unpack
+
+function run_iter(N)
+  local M, ba, bb, buf = 2/N, 2^(N%8+1)-1, 2^(8-N%8), {}
```
vext01: Have you double checked the scaling logic here? It's scary.

vext01: And in the other mandlebrot benchmark.

benchmarks/meteor/bench.lua
```diff
@@ -0,0 +1,227 @@
...
+	  end
+	end
+      end
+    end
+  end
+
+  local s = "local u0,u1,u2,u3,u4,u5,u6,u7,u8,u9" -- Piece use flags.
+  for p=0,99 do if ok[p] then s = s..",b"..p end end -- Board cells.
+  s = s.."\n"..[[
+local countinit = ...
```
vext01: What's the `...` about here?

benchmarks/pidigits/bench.lua
```diff
@@ -0,0 +1,111 @@
...
+function run_iter(N)
+  local pidigit = coroutine.wrap(getdigits)
+  local diget
+  for i=1,N do
+    digit = pidigit()
+  end
+  return digiSt
+end
+
+--[[ Print lines with 10 digits.
```
vext01: Kill commented code?

benchmarks/sum-file.lua
```diff
@@ -0,0 +1,6 @@
+
+local sum = 0
+for line in io.lines() do
+  sum = sum + line
```
vext01: I wasn't expecting that to work! Seems Lua will parse a string looking for a numeric value if you use `+` on it!

```
> print(10 + "11")
21
```

Python:
```
>>> print(10 + "11")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

Today I learned.

iterations_runner.lua
vext01: OK

lualibs/scimarklib_ffi.lua
```diff
@@ -0,0 +1,74 @@
+------------------------------------------------------------------------------
```
vext01: Did you check the license on the libs and benchmarks btw? Do we need to add a license header or file?

fsfod:
@vext01 Most the code is unmodified from the benchmark code i copied it from. The original code was just wrote as a stand-alone script that would just be run directly from the shell and timed. Also i tried to just reuse the existing scaling variable each benchmark had instead creating my own.

benchmarks/mandelbrot/bench.lua
fsfod: Its the original unmodified scaling logic

benchmarks/meteor/bench.lua
fsfod: The top level scope of a script is just another function in Lua that also implicitly takes a variable number of arguments. The ... is how you declare and access the vararg like a tuple of those arguments when you assign it to list of variables it get implicitly unpacked.

vext01:
OK, as long as we have thought about it. Please respond to the other comments, then we can think about squashing.

lualibs/scimarklib_ffi.lua
fsfod: We did already have some of the benchmarks in this repo that came from same repo as all these came from. Mike Paul just has a comment in that repo saying public domain so we just resorted [this before](https://github.com/softdevteam/lua_benchmarking/blob/master/benchmarks/md5/bench.lua#L1)

benchmarks/chamenos/bench.lua
fsfod: Its not really global its declared as a local at the top of the file

benchmarks/sum-file.lua
fsfod: Lua in very early versions started life as something very close to json as i understand it. Also this benchmark i didn't even convert since it just based on reading from stdio really should be removed

vext01: If you think it's not useful, then I'm OK with you killing it. Up to you.

benchmarks/array3d/bench.lua
fsfod: I just used the system it already had

vext01: OK

vext01:
Changes look good. Are you finished? Ready to squash?

fsfod:
These last two commits should be it

vext01:
These look fine, do you want to squash?

vext01:
Are we ready to merge?

fsfod:
@vext01 squashed, but i did copy exact same scaling values for all the new benchmark to the krun file also.

vext01:
I'm leaving those values up to you. Ready to merge?

fsfod:
Yes ready to merge.

PR #44, opened 2018-05-15
Add a benchmark based on the moonscript compiler

fsfod:
Add a large benchmark based on the moonscript compiler that transpiles .moon files to Lua. Currently I'm just compiling the moonscript compiler's own source code as the benchmark. The compiler also has linting mode that could maybe be another benchmark.
Travis also fails on RaptorJIT for moonscript because of the missing shared Lua lib issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,4 @@ script:
   - ./builds/normal/luajit simplerunner.lua --jitstats -c 3
   - ./builds/gc64/luajit simplerunner.lua -e capnproto_decode -c 3
   - ./builds/dualnum/luajit simplerunner.lua -c 3
-  - ./builds/raptorjit/luajit simplerunner.lua --inprocess -e capnproto_decode -c 3
+  - ./builds/raptorjit/luajit simplerunner.lua --inprocess -e capnproto_decode -e moonscript -c 3


ltratt:
How big is this benchmark in KiB? And how many of the files are slurped in verbatim? It would surely be better (from our POV) to download this benchmark in the build script and move files into the right place? Also, what's the benchmark's license?

fsfod:
There is not really a benchmark in moonscript repo i just built one based on very clean API's the compiler was calling [parse.string](https://github.com/leafo/moonscript/blob/master/moonscript/cmd/moonc.lua#L68 ) and [compile.tree](https://github.com/leafo/moonscript/blob/master/moonscript/cmd/moonc.lua#L92).

The source .moon files I'm compiling for the benchmark are only 150kb that will only be about 40kb in git's storage and I had already separated out source files from everything else in the moonscript repo. Also i already fetch the compiler using Lua Rocks.

The licence of the moonscript compiler is MIT like most Lua libraries.

ltratt:
Have any of the `.moon` files been changed at all?

fsfod:
Well they were unmodified from the moonscript repo

fsfod:
@ltratt I squashed it away

fsfod:
@ltratt ready to merge

PR #50, opened 2018-07-11
Add a make file to checkout and build branches of LuaJIT

fsfod:
Add a make file to make building and testing a branch of LuaJIT on the bencher servers less error prone. Also writes some audit files that can used to help check that the binaries being run were created from the right commit.

### DIFF
--- a/ljbuild.make
+++ b/ljbuild.make
@@ -1,0 +1,92 @@
+.SUFFIXES:
+
+ifndef name
+$(error build name is not set)
+endif
+
+BUILD_DIR:= builds/$(name)
+BUILD_FULLDIR:= $(shell cd builds && pwd)/$(name)
+BUILD_DIRS= $(BUILD_DIR) $(BUILD_DIR)/include $(BUILD_DIR)/jit $(BUILD_DIR)/jitlog
+REPO:= $(BUILD_FULLDIR)/src
+SRC:= $(BUILD_DIR)/src/src
+
+# We need a branch name when first checking out the repo
+ifndef branch
+  $(if $(shell if [ -f $(REPO)/.git ]; then echo ok; fi), , $(error No 'branch' variable set))
+endif
+
+# Make sure all output build directories are created before we run anything
+$(shell mkdir -p $(BUILD_DIRS))
+
+SRC_C:= $(wildcard $(SRC)/*.c)
+SRC_H:= $(wildcard $(SRC)/*.h)
+SRC_JITLOG:= $(wildcard $(SRC)/jitlog/*.lua)
+
+# Exclude generated headers from rebuild checking since they break pull_build/reset_build
+GENHDR= lj_bcdef.h lj_ffdef.h lj_libdef.h lj_recdef.h lj_folddef.h  \
+        lj_jitlog_def.h lj_jitlog_writers.h host/buildvm_arch.h
+GENHDR:= $(addprefix $(SRC)/,$(GENHDR))     
+SRC_H:= $(filter-out $(GENHDR),$(SRC_H))
+
+FILES_INC:= luaconf.h lua.h luajit.h lualib.h lauxlib.h
+SRC_INC:= $(addprefix $(SRC)/,$(FILES_INC))
+INCLUDES:= $(addprefix $(BUILD_DIR)/include/,$(FILES_INC))
+
+GITCMD:= git --git-dir="$(REPO)/.git" --work-tree="$(REPO)"
+GITCMD_MAINREPO:= git --git-dir="luajit_repo/.git" --work-tree="luajit_repo"
+
+build: $(BUILD_DIR)/libluajit.so $(BUILD_DIR)/libluajit-5.1.so.2 $(INCLUDES)
+
+$(REPO)/.git:
+	mkdir -p $(REPO)
+	cd luajit_repo && git worktree add --detach $(REPO)
+	$(GITCMD) checkout $(branch)
+	@$(GITCMD) rev-parse --abbrev-ref HEAD > $(BUILD_DIR)/gitbranch.txt
+
+reset: $(REPO)/.git clean
+	$(GITCMD) reset --hard
+
+pull: $(REPO)/.git
+	$(GITCMD) pull
+
+$(BUILD_DIR)/libluajit.so: $(REPO)/.git $(SRC_C) $(SRC_H) $(SRC_JITLOG)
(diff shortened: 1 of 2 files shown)


ljbuild.make
```diff
@@ -0,0 +1,92 @@
+.SECONDEXPANSION:
```
ltratt: What's this doing?

ljbuild.make
```diff
@@ -0,0 +1,92 @@
...
+FILES_INC:= luaconf.h lua.h luajit.h lualib.h lauxlib.h
+SRC_INC:= $(addprefix $(SRC)/,$(FILES_INC))
+INCLUDES:= $(addprefix $(BUILD_DIR)/include/,$(FILES_INC))
+
+GITCMD:= git --git-dir="$(REPO)/.git" --work-tree="$(REPO)"
+GITCMD_MAINREPO:= git --git-dir="luajit_repo/.git" --work-tree="luajit_repo"
+
+$(REPO)/.git:
+	mkdir -p $(REPO)
+	(cd luajit_repo && git worktree add $(REPO))
```
ltratt: I don't think brackets are needed here?

ljbuild.make
```diff
@@ -0,0 +1,92 @@
...
+	rm -f $(BUILD_DIR)/githash.txt
+	rm -f $(BUILD_DIR)/gitstatus.txt
+	rm -f $(BUILD_DIR)/buildflags.txt
+	rm -f $(BUILD_DIR)/build.txt
+	rm -f $(BUILD_DIR)/jit/*.lua
+	rm -f $(BUILD_DIR)/jitlog/*.lua
+	rm -f $(BUILD_DIR)/include/*.h
+	$(MAKE) -C $(SRC) clean 
+
+build: | $(BUILD_DIR)/libluajit.so $(BUILD_DIR)/libluajit-5.1.so.2 $(INCLUDES)
```
ltratt: What does `|` do?

ljbuild.make
```diff
@@ -0,0 +1,92 @@
...
+
+reset_build: | reset clean build
+
+pull_build: | pull clean build
+
+.DEFAULT_GOAL := build
```
ltratt: I don't think this is necesary provided `build` is the first target in the file (which it probably should be).

ljbuild.make
fsfod: Makes the targets after it run in sequential order instead of parallel. It might work without it now but before stuff was running before the code was checked out.

ltratt: That sounds like a bug in the makefile: if those things have to be run in a specific order, there should be a dependency between them?

ljbuild.make
fsfod: I was just running it as a sub shell because it was changing a directory to run the command. I just didn't want the current directory to be left changed if command failed.

ljbuild.make
fsfod: I needed it in the past to delay expansion of variable names but i don't see any $$ variable names left so it can probably go.

ljbuild.make
ltratt: I think each line isolates the CWD anyway?

fsfod: It looks like, i guess i assumed wrong

ljbuild.make
```diff
@@ -0,0 +1,92 @@
+.SECONDEXPANSION:
+.SUFFIXES:
+
+ifndef name
+$(error build name is not set)
```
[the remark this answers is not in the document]
fsfod: I assume it is something built-in. It correctly prints the error for me if i don't assign 'name' on the command line.

ljbuild.make
fsfod: fixed

ljbuild.make
fsfod: fixed

ljbuild.make
fsfod: Fixed moved as first target

ljbuild.make
fsfod: Fixed by adding the the git repo as dependency of header copy and switch 'build' away from sequential

fsfod:
@ltratt I think i answered and fixed all the things you commented on

ljbuild.make
```diff
@@ -0,0 +1,96 @@
...
+REPO:= $(BUILD_FULLDIR)/src
+SRC:= $(BUILD_DIR)/src/src
+
+# We need a branch name when first checking out the repo
+ifndef branch
+  $(if $(shell if [ -f $(REPO)/.git ]; then echo ok; fi), , $(error No 'branch' variable set))
+endif
+
+# Make sure all output build directories are created before we run anything
+$(info $(shell mkdir -p $(BUILD_DIRS)))
```
ltratt: What does `info` do?

ljbuild.make
```diff
@@ -0,0 +1,96 @@
...
+	rm -f $(BUILD_DIR)/githash.txt
+	rm -f $(BUILD_DIR)/gitstatus.txt
+	rm -f $(BUILD_DIR)/buildflags.txt
+	rm -f $(BUILD_DIR)/build.txt
+	rm -f $(BUILD_DIR)/jit/*.lua
+	rm -f $(BUILD_DIR)/jitlog/*.lua
+	rm -f $(BUILD_DIR)/include/*.h
+	$(MAKE) -C $(SRC) clean 
+
+reset_build: | reset clean build
```
ltratt: I think we should delete this target and the one below: people should do these steps manually; and we'll forget these targets are here (so they'll just clog things up).

ljbuild.make
fsfod: It it just prints text to the console idk why i left it wrapped in that. I know that the code was a loop before i simplified it at least. It should be fixed now

ljbuild.make
fsfod: I was using them on the benchers to make sure when i fixed something that it was building the latest commit. I did originally have the 'build' target always try to pull first as well.

ltratt: On the bencher just do `make reset clean build` or equivalent. I'm just never keen on these meta-rules: they're rarely used, and thus subject to bitrot.

fsfod: It needs to be ordered

fsfod:
@ltratt I fixed the last two issues.

ltratt:
Works for me. Please squash.

fsfod:

PR #51, opened 2018-07-25
Make VM definitions less verbose and more flexible in the krun file

fsfod:
Simplified the VM definitions for the various LuaJIT build flavors to just to just a single line
Add supporting for specifying per VM environmental variables.

### DIFF
--- a/luajit.krun
+++ b/luajit.krun
@@ -14,12 +14,6 @@ DIR = os.getcwd()
 
 VM_ENV = '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(os.path.join(DIR, "lualibs"))
 
-def make_env(vmdir):
-  basedir = os.path.join(DIR, "builds", vmdir)
-  return {
-    'LUA_PATH' : VM_ENV + '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(basedir)
-  }
-
 HEAP_LIMIT = 2097152  # KiB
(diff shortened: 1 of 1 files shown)


ltratt:
I think we need to make the indents 4 spaces? Otherwise it looks fine.

fsfod:
@ltratt I fixed the indent if you want to merge.

luajit.krun
```diff
@@ -32,6 +26,23 @@ ITERATIONS_ALL_VMS = 1500
 ITERATIONS_NO_JIT = 100
 INSTRUMENT = False
 
+def make_env(vmdir, extra=None):
+    basedir = os.path.join(DIR, "builds", vmdir)
+    ret = {
+        'LUA_PATH' : VM_ENV + '"{0}/?/init.lua;{0}/?.lua;{0}/?/?.lua;"'.format(basedir)
+    }
+    if extra != None:
```
ltratt: Should be `if extra is not None`.

luajit.krun
```diff
@@ -32,6 +26,23 @@ ITERATIONS_ALL_VMS = 1500
...
+        ret.update(extra)
+    return ret
+
+def makevmdef(dirname, instrument=False, env_extra=None, n_iterations=None):
+    ret = {
```
ltratt: Should just be `return { ... }?` and drop the `ret =`?

fsfod:
@ltratt I fixed both those issues.

ltratt:
Please squash.

fsfod:

PR #52, opened 2018-08-07
Update Raptorjit commit lock

fsfod:
Fix the broken LuaJIT commit lock and update the raptorjit one to  v1.0.0alpha1 that was just recently released. The main build script was changed to abort if any of it commands return non zero.
Also exclude the moonscript benchmark for raptorjit since there is no lua51.so built by raptorjit needed for the binary Lua modules used by the benchmark.

### DIFF
--- a/luajit.krun
+++ b/luajit.krun
@@ -125,6 +125,8 @@ BENCHMARKS = {
 SKIP = [
     "capnproto_decode:GC64:*", # Sometime crashes with GC64
     "capnproto_decode:RaptorJIT:*", # Sometimes crashes as well
+	# We can't use binary Lua modules with Raptorjit because it statically links the Lua shared library
+    "moonscript:RaptorJIT:*",
     #"*:Normal:*",
     #"*:GC64:*",
     #"*:NoJIT:*",
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Make sure we bail out if any of our commands fail, like the commit lock
+set -e
+
 #Build the various interesting flavours of LuaJIT
 if [ $# -gt 1 ]; then
   echo "Usage: build.sh [<LuaJIT source dir>]" 2>&1
--- a/commitlock.txt
+++ b/commitlock.txt
@@ -1,2 +1,2 @@
-luajit_commitid=655d8f29ae0a3c4cd1c57c6c8c88e6be41aed21c
+luajit_commitid=a6e64201a043c397f43f10eab9fef2f02adf88e5
 raptorjit_commitid=39680ca7ebc3ce43f68c465bedb3e88f074aaeea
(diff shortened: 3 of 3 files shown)


ltratt:
Tests are failing?

fsfod:
Well it looks like raptorjit is getting stuck in an infinite loop or something when JITing loops

ltratt:
Can you file a bug against RaptorJIT for this please?

luajit.krun
```diff
@@ -125,6 +125,7 @@ BENCHMARKS = {
 SKIP = [
     "capnproto_decode:GC64:*", # Sometime crashes with GC64
     "capnproto_decode:RaptorJIT:*", # Sometimes crashes as well
+    "moonscript:RaptorJIT:*",
```
ltratt: Why are we skipping this?

fsfod: RaptorJIT is statically linking the Lua shared library and binary Lua modules are linked to it as shared a library. Moonscript happens to use some binary Lua modules

fsfod:
@ltratt The strange thing is it doesn't seem to get in an infinite loop when run under krun

ltratt:
That sounds weird. Any thoughts?

fsfod:
@ltratt Well its looks it getting stuck in an infinite loop trying to log new ffi ctypes when a trace ends. I guess no good deed goes unpunished because that code is based on something i suggested to Luke.

fsfod:
Well now its getting stuck in the fasta benchmark after about 50 iterations

fsfod:
So RaptorJIT is leaking tons of memory running fasta and i think it be might be getting in to state where it can't progress forward from trying to allocate a new machine code arena to write a completed a trace in to but just aborts the trace when it can't and keeps doing this.

luajit.krun
```diff
@@ -125,6 +125,9 @@ BENCHMARKS = {
 SKIP = [
     "capnproto_decode:GC64:*", # Sometime crashes with GC64
     "capnproto_decode:RaptorJIT:*", # Sometimes crashes as well
+	# We Can't use binary Lua modules with raptojit because it statically
```
ltratt: "raptoRjit"

ltratt: [Also, the sentence as a whole doesn't parse.]

fsfod:
@ltratt I've tried to improve the comment and i'm now using the memory leak fix Luke made for Raptojit.

ltratt:
Please squash.

fsfod:
@ltratt Squashed and I switched back to the main Raptorjit repo since both the memory leak and infinite loop fixes are merged in now.

ltratt:
Where's the commit with that update?

fsfod:
well it was squashed but i put it back

fsfod:

ltratt:
Thanks!

ltratt (approved):

